PR #3, opened 2018-12-19
Add new event

k3ntako:
Added Font Awesome and React-Flatpickr


client/models/Location.js
```diff
@@ -14,19 +14,32 @@ const FAKE_LOCATION = {
...
+        address: "Massachusetts Ave Bridge, Cambridge, MA 02142",
+        note: "Make sure to dropoff your bags in the lockers before meeting."
       }
     }
   }
 }
 
 export default class Location extends Model {
 
+  static allLocations(){
```
mienaikoe: This list can get really big once we have multiple teams.

Make this function take in a team id, so it's `listByTeam( teamId )`

Also, you don't need the word "Locations" in the function name because the way you would call it is `Location.fxn()`

client/models/Person.js
```diff
@@ -20,6 +20,10 @@ const FAKE_USER = {
 
 export default class Person extends Model {
 
+  getCurrentUser(){
```
mienaikoe: function naming: You don't need the word "User". See above for justification

client/pages/AddEvent/AddEventDropdown.js
```diff
@@ -0,0 +1,21 @@
+import React, {Component} from 'react';
+
+
+export default class AddEventDropdown extends Component {
+  render(){
+    const boatOptions = this.props.boats.map(boat => {
```
mienaikoe: James never has set boat loadouts - there was always a new permutation for each practice or race. It doesn't make sense to select one from before.

client/pages/AddEvent/AddEventDropdown.js
```diff
@@ -0,0 +1,21 @@
+import React, {Component} from 'react';
+
+
+export default class AddEventDropdown extends Component {
```
mienaikoe: Also, this is a boat dropdown, but you don't state this in the class name

client/pages/AddEvent/index.js
```diff
@@ -0,0 +1,136 @@
...
+import Boat from '../../models/Boat';
+
+import AddEventDropdown from './AddEventDropdown';
+
+export default class AddEventPage extends Component {
+
+  constructor(props){
+    super(props);
+    this.state = {
+      locations: [],
```
mienaikoe: the list of the team's locations shouldn't be in the state. This is an immutable object. Ideally this would be a prop from redux, but redux would be overly much.

client/pages/AddEvent/index.js
```diff
@@ -0,0 +1,136 @@
...
+      name: "",
+      date: new Date(),
+      locationId: "default",
+      boats: [],
```
mienaikoe: same as above, nix both lines 20 and 21. You can handle boat loadouts once the event is created - will make everything 10x easier

client/pages/AddEvent/index.js
```diff
@@ -0,0 +1,136 @@
...
+      }else if(boats[i] === boats[i-1]){
+        return alert("Please select a different boat for each or change the number of boats.");
+      }
+    };
+
+    date = date.valueOf();
+
+    if(name.replace(/\s/g,'') && date && locationId !== "default"){
+      Event.create(this.state.name, date, locationId);
+      this.props.history.push(`/`);
```
mienaikoe: 1) You should wait until the event is created. use Event.create(...).then( () => {
});

2) You shouldn't push them back to the homepage. The user should have feedback that their event has been created, so you should take them to the event page.

client/pages/AddEvent/index.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }
+
+  componentDidMount(){
+    Location.allLocations().then( locations => {
+      this.setState({locations});
+    });
+
+    //get all boats for team
+    this.setState({boats: [{id: 1, name: "Black Boat"}, {id: 2, name: "Red Boat"}]});
```
mienaikoe: nix

client/pages/AddEvent/index.js
```diff
@@ -0,0 +1,136 @@
...
+    if(name.replace(/\s/g,'') && date && locationId !== "default"){
+      Event.create(this.state.name, date, locationId);
+      this.props.history.push(`/`);
+    }else{
+      alert("Could not add event. Please make sure all fields are filled out.");
+    }
+  }
+
+  componentDidMount(){
+    Location.allLocations().then( locations => {
```
mienaikoe: as before, fetch by team id

client/pages/Events/index.js
```diff
@@ -1,32 +1,39 @@
...
+      this.setState({ events });
     })
   }
 
   render(){
     let eventsHtml;
     if(this.state.events){
       eventsHtml = this.state.events.map(event => {
-        return <EventCard key={event.id} event={event} />
+        return <EventCard key={event.id} event={event} clickHandler={this.cardClickHandler} />
```
mienaikoe: You should pass in a link url. A click handler won't reflect that an item is a link in the browser, and make people confused about what a card does or whether it's clickable.

client/pages/AddEvent/AddEventDropdown.js
k3ntako: I was thinking that he can at least select the name of the boat. But, I'm realizing that he won't know the number of boats required until after people respond. I will just take out the boats all together in add event.

client/pages/AddEvent/index.js
k3ntako: For #2, what should I do for now? We can't really take them to a new page since we can't add new events?

client/models/Person.js
k3ntako: Removed it because I'm not actually using the current user.

client/pages/AddEvent/AddEventDropdown.js
k3ntako: Deleted dropdown.

client/pages/Events/index.js
mienaikoe: The latter.

client/pages/AddEvent/index.js
mienaikoe: Just take them to `event/:id`

client/pages/AddEvent/index.js
k3ntako: Jesse will set up providers and consumers.

PR #4, opened 2018-12-21
Edit an event

k3ntako:
Need to change alert style.
Not clear that date is editable. 
Returned event initially returns a location object. However, after updating the returned location is in an array. This causes an issue where location is not displayed after being updated.


client/pages/Event/index.js
```diff
@@ -1,20 +1,37 @@
...
+import Location from '../../models/Location';
 import EventFormatter from '../../formatters/EventFormatter';
 
 export default class EventPage extends Component {
 
   constructor(props){
     super(props);
-    this.state = {};
+    this.state = {
+      editEvent: false
```
k3ntako: change to isEditing

client/pages/Event/index.js
```diff
@@ -23,7 +40,43 @@ export default class EventPage extends Component {
...
+  dateChange(datetime){
+    this.setState({datetime});
+  }
+
+  toggleEdit(){
+    if(this.state.editEvent){
+      let {name, datetime} = this.state;
+      datetime = datetime.valueOf();
+      const location = this.state.locations.filter(loc => loc.id === this.state.locationId);
+      this.state.event.update(name, datetime, location).then(() => {
```
k3ntako: replace `location` with `locationId`

client/pages/Event/index.js
```diff
@@ -1,20 +1,37 @@
...
+    this.locationChangeHandler = this.locationChangeHandler.bind(this);
   }
 
   componentDidMount(){
     Event.find(this.props.match.params.id).then( event => {
+      const { name, datetime, location } = event;
       this.setState({
+        name,
+        datetime,
+        locationId: location.id,
```
k3ntako: put this in another object. `updateParams`

k3ntako: User `event` to show on screen

client/pages/Event/index.js
k3ntako: have a cancel option. 
`setState` at least twice: turning on and off edit mode.
Make sure edit and save buttons are in different locations in case of double click.

client/pages/Event/index.js
```diff
@@ -32,31 +85,59 @@ export default class EventPage extends Component {
...
-        </div>
-        <div className="boats">
-          <h2>No Answer - {this.state.event.no_answers.length}</h2>
-          {noAnswers}
-          <h2>Attendees - {this.state.event.attendees.length}</h2>
-          {attendees}
-          <h2>Not Attending - {this.state.event.non_attendees.length}</h2>
-          {nonAttendees}
-        </div>
+      {title}
```
k3ntako: Wrap in `h1`.
Make sure user can highlight text.

k3ntako:
Issue with log in `TypeError: Cannot read property 'map' of undefined`. Clicking `Submit` again returns `Error: You're already logged in`.

k3ntako:
I meant the fake attendance data that we were displaying on the event page. Haven't looked into it, but each person was undefined.

mienaikoe:
Fixed it

PR #5, opened 2019-01-01
Boat page

k3ntako:


client/pages/Boat/AttendeesDropdown.js
```diff
@@ -0,0 +1,28 @@
+import React, {Component} from 'react';
+
+
+export default class AttendeesDropdown extends Component {
```
mienaikoe: In cases of having a list that is greater than 10 items I generally go for an autocomplete instead of a dropdown. Scrolling through 50 items to find the one you're looking for is cumbersome.

client/pages/Boat/index.js
```diff
@@ -0,0 +1,82 @@
...
+      this.setState({ boat });
+    });
+
+    Event.find(1).then( event => {
+      this.setState({ event });
+    });
+  }
+
+  makeListOfSeats(boat){
+    return boat.people.map((person, idx) => {
```
mienaikoe: in the new schema, it's boat.seats, and each seat has a "seat name" and a "person"

package.json
```diff
@@ -11,6 +11,8 @@
   "license": "ISC",
   "dependencies": {
     "connect-mongo": "^2.0.3",
+    "dotenv": "^6.2.0",
+    "dotenv-webpack": "^1.5.7",
```
mienaikoe: looked this up. looks safe.

public/index.css
```diff
@@ -2,8 +2,16 @@ body{
   padding: 0;
   margin: 0;
   font-size: 12px;
-  font-family: sans-serif;
   background-color: #eee;
+  font-family: 'Comfortaa', cursive;
```
mienaikoe: cursive should never be used as a fallback font. It's illegible and has irregular rendering width.

client/pages/Boat/AttendeesDropdown.js
mienaikoe: That, said, could totally address that in a different commit

k3ntako: Yep I will do this another commit.

public/index.css
```diff
@@ -140,15 +140,20 @@ nav{
 }
 
 .add-event-button{
```
mienaikoe: I know you want to customize this button with animations, but let's stick to whatever comes out of the box from react-toolbox for now, and get fancy a bit later.

mienaikoe:
not sure what you did, but i was able to merge successfully...

PR #6, opened 2019-01-07
Event response

k3ntako:
Fixed back end issues:
1) `attendees` and `non_attendees` were coming back as `undefined` instead of `Person` objects.
2) If user clicked on the same response ('Attending' or 'Skipping') multiple times, their response would be recorded multiple times in the database.

Issues with this PR:
1) Did not implement `MeConsumer`. Used `Person.me()` instead. 
2) Could not utilize `react-toolbox-themr`.


client/models/Event.js
```diff
@@ -41,6 +41,7 @@ class Event extends Model {
       } else {
         this.non_attendees.push(Person.currentUser);
       }
+      return this;
```
mienaikoe: Oh.. that was obvious. lol

client/models/Model.js
```diff
@@ -53,7 +53,7 @@ class Schema {
       if( field.subtype ){
         if( field.type == Array ){
           ret[field.key] = objVal.map( subObj => {
-            new field.subtype(subObj);
+            return new field.subtype(subObj);
```
mienaikoe: another obvious.. jeez. i need to stop coding at midnight.

client/pages/Event/index.js
```diff
@@ -1,25 +1,53 @@
...
+          if(not_attending){
+            myResponse = "not_attending";
+          }
+
+        }
+
+        return myResponse
+
+      }).then(myResponse => {
+        this.setState({ event, myResponse });
```
mienaikoe: never thought of doing this - nesting promises means you can use the params from both.

client/pages/Event/index.js
```diff
@@ -33,18 +61,41 @@ export default class EventPage extends Component {
     });
   }
 
+  attendingClickHandler(){
+    this.state.event.attend(true).then(event => {
+      this.setState({
+        event,
+        myResponse: "attending"
+      });
+    });
```
mienaikoe: I would add some error handling here.

client/pages/Event/index.js
```diff
@@ -33,18 +61,41 @@ export default class EventPage extends Component {
...
+  }
+
+  notAttendingClickHandler(){
+    this.state.event.attend(false).then(event => {
+      this.setState({
+        event,
+        myResponse: "not_attending"
+      });
+    });
```
mienaikoe: and here

client/pages/Event/index.js
```diff
@@ -62,6 +113,10 @@ export default class EventPage extends Component {
           <h2>Boats</h2>
           {boats}
         </div>
+        <div className="event-page-section">
+          <Button label='Attending' className={styles.attendButton} raised onMouseUp={this.attendingClickHandler} disabled={disableAttending}/>
+          <Button label='Skipping' className={styles.notAttendButton} raised onMouseUp={this.notAttendingClickHandler} disabled={disableNotAttending}/>
```
mienaikoe: People usually use `onClick` instead of `onMouseUp`

server/controllers/EventController.js
```diff
@@ -61,24 +61,27 @@ const EventController = {
...
         $pull: {
-          attendees: [req.session.person_id]
+          attendees: req.session.person_id
         }
       };
+
     return Event.updateOne({
-      _id: req.params.event_id
+      _id: req.params.event_id,
+      [addingTo]: {$ne: req.session.person_id}
```
mienaikoe: This shouldn't be necessary

server/controllers/EventController.js
```diff
@@ -61,24 +61,27 @@ const EventController = {
...
+          non_attendees: req.session.person_id
         }
       } : {
         $push: {
-          non_attendees: [req.session.person_id]
+          non_attendees: req.session.person_id
         },
         $pull: {
-          attendees: [req.session.person_id]
+          attendees: req.session.person_id
```
mienaikoe: have you run tests on this?

server/controllers/EventController.js
k3ntako: Without that line, if I click on Attending or Skipping multiple times, it adds to the database multiple times. 

Right now removing that line won't do add a person multiple times, because I disable the button in React, but I think a second line of defense for bad data isn't bad.

mienaikoe: ahh I had thought I was doing a pull from both lists. nvm my comment - this is a better way to do this.

server/controllers/EventController.js
k3ntako: Not exactly sure what you mean by tests (I didn't write tests). I coincidentally learned about `$push`/`$pull` couple hours before I saw your code, so I don't know much about them. 

I played around with/without the brackets. From my experience, `$pull` doesn't seem to work when I have brackets, but `$push` does. I end up with a copy of me on both `attending` and `non_attending`. 

I'm looking at this example from the docs on `$pull`. In **Update**, it seems that "carrots" isn't in brackets, and "apples" and "oranges" are but require `$in`.
https://docs.mongodb.com/manual/reference/operator/update/pull/

*The following operation updates all documents in the collection to remove "apples" and "oranges" from the array fruits and remove "carrots" from the array vegetables:*

**Original Data:**
```
{
   _id: 1,
   fruits: [ "apples", "pears", "oranges", "grapes", "bananas" ],
   vegetables: [ "carrots", "celery", "squash", "carrots" ]
}
{
   _id: 2,
   fruits: [ "plums", "kiwis", "oranges", "bananas", "apples" ],
   vegetables: [ "broccoli", "zucchini", "carrots", "onions" ]
}
```
**Update:**
```
db.stores.update(
    { },
    { $pull: { fruits: { $in: [ "apples", "oranges" ] }, vegetables: "carrots" } },
    { multi: true }
)
```
**Result:**
```
{
  "_id" : 1,
  "fruits" : [ "pears", "grapes", "bananas" ],
  "vegetables" : [ "celery", "squash" ]
}
{
  "_id" : 2,
  "fruits" : [ "plums", "kiwis", "bananas" ],
  "vegetables" : [ "broccoli", "zucchini", "onions" ]
}
```

`$push` seems to be similar. If you want to add multiple values in brackets, they require `$each`.
https://docs.mongodb.com/manual/reference/operator/update/push/

mienaikoe: huh TIL.

Just run `npm run test` to make sure this works.

k3ntako: 17 passing!

client/pages/Event/index.js
mienaikoe: actually - too early to do this.

client/pages/Event/index.js
mienaikoe: too early to do this

PR #10, opened 2019-01-15
Boat page

k3ntako:


client/models/Boat.js
```diff
@@ -37,15 +37,10 @@ class Boat extends Model {
...
       user_id: user._id,
-      seat: seat
-    }).then( () => {
-      this.seats.push({
-        user: user._id,
-        seat: seat
-      });
-    });
+      seat_name: seat
+    }).then( boatJson => new Boat(boatJson) );
```
mienaikoe: I think it's wasteful and risks bugs to return the entire boat back from this call.

client/pages/Boat/index.css
```diff
@@ -0,0 +1,39 @@
+.seatsContainer{
+  width: calc(78vw + 30px);
+  margin: 0 auto 30px auto;
+
+  display: grid;
+  grid-template-columns: 20vw 19vw 40px 19vw 20vw;
```
mienaikoe: word of warning: vw and vh units are rare - you might not actually want to do this, and do something responsive with a maximum width instead. The reason is because you can't control the size of the viewport, it will look not so great for people with giant screens or tablets that are just over the responsive breakpoint.

That said, we can address this in Layout + Design.

client/pages/Boat/index.js
```diff
@@ -20,61 +38,78 @@ export default class BoatPage extends Component {
...
-        <AttendeesDropdown
-          attendees={this.state.event.attendees}
-          assignedAttendeeId={assignedAttendeeId}
-          seatId={idx}
-          handleSeatChange={this.handleSeatChange}
+      const searchText =  user ? user.name : "";
+      const inputClassName = user ? styles.seat + " " + styles.seatAssigned : styles.seat;
+
+      let position = "paddler";
+      position = seatLabel === "D" ? `drummer ${String.fromCodePoint("129345")}` : position;
```
mienaikoe: What is `${String.fromCodePoint("129345")}` for?

Edit: You can use emojis in your code. Just be aware that people that are running Windows 7 or have mac 10.11 or lower will not be able to see emojis.

Therefore, I highly suggest you don't use emoji. Let's use an icon instead or get an image from the interwebs.

client/pages/Boat/index.js
```diff
@@ -20,61 +38,78 @@ export default class BoatPage extends Component {
...
-    }else{
-      this.state.boat.setSeat(seatId, attendee).then(boat => {
-        this.setState({ boat });
-      })
-    }
+  handleSeatChange(val, seatId){
+    const attendee = this.state.event.attendees.find(attendee =>  attendee.name === val);
+
+    this.state.boat.seat(attendee, seatId).then(boat => {
+      this.setState({ boat });
```
mienaikoe: so like I said previously, it's better to change the existing boat object than to replace the entire boat object. It's expensive both in IO, processing speed, and memory to send an object over the wire when a simple 200 will do.

client/pages/Event/AddBoat.js
```diff
@@ -0,0 +1,88 @@
...
+
+  addBoat(){
+    const newBoatName = this.state.newBoatName === null ? "" : null;
+    this.setState({ newBoatName });
+  }
+
+  saveBoat(){
+    let newBoatName = this.state.newBoatName;
+    let event = this.props.event;
+    if(newBoatName.replace(/\s/g,'')){
```
mienaikoe: handle this in `boatOnChangeHandler` so that the user understands what's going on.

Also why remove white space? That seems heavy-handed and unnecessary.

client/pages/Event/AddBoat.js
```diff
@@ -0,0 +1,88 @@
...
+    let saveBoatButton;
+    if(this.state.newBoatName !== null){
+      newBoatInput = <Input
+        type="text"
+        value={this.state.newBoatName}
+        onChange={this.boatOnChangeHandler}
+        placeholder="Boat Name"
+        />;
+      addBoatButtonText = "Cancel";
+      saveBoatButton = <Button label='Save Boat' raised onClick={this.saveBoat}/>;
```
mienaikoe: disable the button if there's no text in the name input

client/pages/Event/index.css
```diff
@@ -1,7 +1,27 @@
+.eventPageSection{
+  padding: 1.17em;
```
mienaikoe: units all over the place. pick em or px or rem. Don't mix them up unless there's a good reason to do so.

client/pages/Event/index.js
```diff
@@ -14,12 +19,15 @@ export default class EventPage extends Component {
     this.state = {
       myResponse: "no_response"
     };
+
+    this.eventId = this.props.match.params.id;
```
mienaikoe: What if you go from one event page to another? You should avoid saving variables directly onto a React component

server/controllers/BoatController.js
```diff
@@ -52,35 +52,37 @@ const BoatController = {
...
+            ]
+          }
         }
-        return boat.save();
-      }).
-      then( aBoat => {
+      } : {
+        $pull: {
+          seats: {
+            person: req.body.user_id
```
mienaikoe: this seems redundant to the previous. Something fishy is going on here but I'll look into it after the commit

server/controllers/BoatController.js
```diff
@@ -52,35 +52,37 @@ const BoatController = {
...
+      };
+
+    return Boat.findByIdAndUpdate(req.params.boat_id, boatUpdater, {
+      new: true
+    }).
+      then( boat => {
         if( req.body.seat_name ){
           boat.seats.push({
-            person: req.body.person_id,
+            person: req.body.user_id,
```
mienaikoe: ugh didn't realize I called it user_id.... should be called person_id. I'll handle after merge.

client/pages/Boat/index.js
k3ntako: Yep, I will take this out! 😛

client/models/Boat.js
```diff
@@ -37,15 +37,10 @@ class Boat extends Model {
   }
 
   seat(user, seat){
-    return Fetcher.post(`api/boats/${this._id}/seat`,{
+    return Fetcher.post(`/api/boats/${this._id}/seat`,{
       user_id: user._id,
-      seat: seat
-    }).then( () => {
-      this.seats.push({
-        user: user._id,
```
k3ntako: Key should name and person. And user should be person, not user._id

client/pages/Event/AddBoat.js
k3ntako: Use trim here.

client/pages/Event/index.js
k3ntako: just use variables

PR #11, opened 2019-01-16
Styling of events list

mienaikoe:
First pass at Layout & Design of Events List page. Includes responsive design and a new colors retheme. Also includes a fix on `client/models/Model`


client/models/Model.js
```diff
@@ -5,12 +5,13 @@ class Model {
   constructor(props){
     if( !props ){
       return null;
-    } else if( props instanceof String ){
-      this._id = props;
+    } else if( typeof props === 'string' ){
+      this._id = props; // this is really dangerous
```
k3ntako: Can we use `Object.isValid()` to check? 
https://mongodb.github.io/node-mongodb-native/api-bson-generated/objectid.html#objectid-isvalid
https://stackoverflow.com/questions/13850819/can-i-determine-if-a-string-is-a-mongodb-objectid

mienaikoe: nah once it gets to the client side, it's not a mongo object anymore - just a plain json object

client/models/Event.js
```diff
@@ -45,6 +45,14 @@ class Event extends Model {
     })
   }
 
+  is_attending( person ){
+    if( !this.attendees.length ){
+      return false;
+    }
+
+    return !!this.attendees.find( attendee => attendee._id === person._id );
```
k3ntako: Wouldn't this return false if length is 0? If so, you wouldn't need the if statement above.
I would imagine that it wouldn't take much computational power to loop through an empty array.

mienaikoe: Oh.. yea that makes sense. Will fix.

client/pages/Events/EventCard.css
```diff
@@ -0,0 +1,24 @@
+
+.datetime{
+  font-size: 1.6rem;
+  margin: 1rem;
+}
+
+.root{
+  
+}
```
k3ntako: Are you planning to add more to `.root`?

mienaikoe: possibly? I added this in case I needed it in the future. If I remove it, I think the css would resolve `styles.root` to undefined, and then the className will be "undefined"

PR #14, opened 2019-01-20
Adding Person.listForMyTeam

mienaikoe:
Lists all the people on "my" team.


client/models/Person.js
```diff
@@ -75,6 +75,12 @@ class Person extends Model {
     });
   }
 
+	static listForMyTeam(){
+		return Fetcher.get('/api/people/teammates').then( people => {
+			return people.map( person => new Person(person) );
+		})
+	}
+
```
k3ntako: Seems like you indented more above than below.

server/controllers/PersonController.js
```diff
@@ -164,6 +164,23 @@ const PersonController = {
...
+				});
+			}).
+			then( people => {
+				res.json(people);
+				next();
+			}).catch( err => {
+				next(err);
+			})
+	},
+
```
k3ntako: Same as above, more indenting.

PR #15, opened 2019-01-21
Set a Boat as Visible/Invisible

mienaikoe:


mienaikoe:
Yea, but figured that could be done in the frontend changes

PR #16, opened 2019-01-21
Change seat assignment to dialog

k3ntako:
Addressed bug in React Toolbox. Might need to delete the `node_module` and reinstall.

**Changes:**
https://github.com/k3ntako/react-toolbox/commit/e0464bf582c928d75b15efbdaf36ae7efcad1b09

**Issue:**
https://github.com/react-toolbox/react-toolbox/issues/1900


mienaikoe:
Be sure to leave a not about your PR in the linked Issue

mienaikoe:
Also, instead of requiring other devs to reinstall node_modules, you should make a tag of the state of that other repo and then reference that tag in package.json

k3ntako:
Thanks for the feedback! I tagged it and updated package.json.

client/pages/Boat/SeatPicker.js
```diff
@@ -0,0 +1,72 @@
...
+    this.teamList = this.teamList.bind(this);
+  };
+
+  componentDidMount(){
+    this.teamList();
+  }
+
+  teamList(){
+    const teamList = this.props.team.map(person => person.name);
+    this.setState({ teamList });
```
mienaikoe: Since this state variable is dependent on props, you should also call this function in `componentWillReceiveProps`

mienaikoe: Also, you could sort it by preferred position and preferred side - sorting by which of these fields the calling component wants shown first.

You can also sort by who's actually going and who's not. And also you should render some notification to know all of these factors that go into deciding who to put in a seat.

alas I'm getting ahead of myself and react-toolbox doesn't have something like that out of the box. Let's handle it next pass.

client/pages/Boat/index.js
```diff
@@ -28,58 +30,61 @@ export default class BoatPage extends Component {
...
     });
 
+    Person.listForMyTeam().then(team => {
+      this.setState({ team })
+    })
+
     Event.find(this.props.match.params.eventId).then( event => {
-      this.setState({ event }, this.listOfAttendees);
+      this.setState({ event });
     });
```
mienaikoe: You can combine these promises and then do a single setState call, which reduces render passes. Lookup `Promise.all`

mienaikoe: I also think you could add `Person.listForMyTeam` as a call that happens once someone logs in or joins so you don't have to keep calling it everytime someone goes to the boat page. You can wrap that in a provider/consumer pattern so you can put it anywhere you want.

package.json
```diff
@@ -21,7 +21,7 @@
     "react-google-maps": "^9.4.5",
     "react-router": "^4.3.1",
     "react-router-dom": "^4.3.1",
-    "react-toolbox": "github:k3ntako/react-toolbox#BugFixes",
+    "react-toolbox": "github:k3ntako/react-toolbox#2.0.0-beta.13-k1.0.3",
```
mienaikoe: Lol this is the longest tag string i've ever seen. [Just observing, you can resolve this]

client/pages/Boat/SeatPicker.js
k3ntako: `componentWillReceiveProps` seems to be deprecated. I used `static getDerivedStateFromProps()` instead for the map. Any reason not to use that?
https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

Okay, so I will do some sorting. I can at least change the order based on the seat selected.
Not 100% sure what you mean by the notification. I did put `preferred_seat`, `preferred_side`, and `weight` on the dialog used for assigning seats.

client/pages/Boat/index.js
k3ntako: Yep will use `Promise.all`.

Just wanna make sure I'm understanding this. I will create a provider/consumer that will hold the one and only team we currently have. But, in the future, if we choose to add more teams, we will need to change the provider/consumer and `Person.listForMyTeam`?

client/pages/Boat/SeatPicker.js
k3ntako: I'm realizing that `static getDerivedStateFromProps()` won't be able to tell when the team changed. Without a team_id or something to keep track, this won't work.

k3ntako: I ended up using `static getDerivedStateFromProps()`.

Sorting seems harder than expected. I do have an idea for another way of doing it that might be better. I will explore it later today.

mienaikoe: Yea - I thought about it more, and I think those really should be suggested people that might be outside the the autocomplete.

getDerivedStateFromProps is fine. Let's handle sorting on second pass.

client/pages/Boat/SeatPicker.js
```diff
@@ -0,0 +1,90 @@
...
+
+export default class SeatPicker extends Component {
+
+  constructor(props){
+    super(props);
+    this.state = {}
+  };
+
+  static getDerivedStateFromProps(props){
+    function sortBySide(a,b,side){
```
mienaikoe: this function should be a scope level up so you don't redefine the function every time you call getDerivedState

client/pages/Boat/index.js
mienaikoe: You've reused variables here - `let boat` and `const boat` might be ok with the compiler, but it's a very bad code practice to use the same name for multiple variables because it confuses the reader.

k3ntako: So, one account can be only associated with one team?

Definitely didn't mean to do that.

client/pages/Boat/SeatPicker.js
k3ntako: I fixed it!

client/pages/Boat/index.js
mienaikoe: Not necessarily. Maybe in future iterations, you choose your team from the get-go. Or maybe we fetch all people from all of your teams and bucket them. Either way I think it's too early to think about that - let's just make it for one season of msk and see if it gets used / how it gets used.

mienaikoe: Looks like you used Promise.all. Resolving

PR #20, opened 2019-01-25
Event - fix and refactor

k3ntako:


client/models/Event.js
```diff
@@ -81,10 +81,12 @@ Event.schema = new Model.Schema([
...
   },{
     key: "datetime",
-    type: Date
+    type: Date,
+    subtype: Date
   },{
     key: "location",
-    type: Location
+    type: Location,
+    subtype: Location
```
k3ntako: Without a subtype it returns an object, not a Location instance

mienaikoe: Something is fishy about that. Location is a first-level child of Event

mienaikoe: Oh jeez another problem with the Model class. I'm going to add another commit to this branch

PR #22, opened 2019-01-27
big fix to a bunch of things

mienaikoe:
I realized:
- that frontend models weren't working
- the team should have a list of members for convenience and future proofing

After this gets committed, you should run the migration.


client/models/Boat.js
```diff
@@ -70,7 +70,7 @@ class Boat extends Model {
 
 }
 
-Boat.schema = new Model.Schema([
+Model.registerFields(Boat, [
```
mienaikoe: Found a problem with circular dependencies, so I had to rejigger the way a schema gets created and declared.

server/models/Person.js
```diff
@@ -3,11 +3,11 @@ const mongoose = require('mongoose')
 const ObjectId = mongoose.Schema.Types.ObjectId;
 
 const PREFERRED_SIDES = [
-  "L","R"
+  "L","R",null
```
mienaikoe: apparently enums have to include null if the field can be nullable

server/models/Team.js
```diff
@@ -7,7 +7,10 @@ const Team = mongoose.model('Team', {
   name: { type: String, unique: true },
 	coaches: [
 		{ type: ObjectId, ref: 'Person' }
-	]
+	],
+  members: [
```
mienaikoe: This is the right way to specify a many-to-many relationship

server/models/Person.js
```diff
@@ -3,11 +3,11 @@ const mongoose = require('mongoose')
...
 ];
 
 const PREFERRED_POSITIONS = [
-  "P","E","R","D","S"
+  "P","E","R","D","S",null
```
k3ntako: I assume P is the section of pacers (rows 1-3). Should we have something for the first row? Generally, coach sticks to a few people for the first row during races.

Also, what is the reason for using single letters opposed to words (Pacers, Engine...)? Might be clearer that way?

server/controllers/TeamController.js
```diff
@@ -0,0 +1,43 @@
+const mongoose = require('mongoose')
+
+const { ValidationError, ForbiddenError, NotFoundError } = require('../errors');
+const Person = require('../models/Person')
+const Team = require('../models/Team')
+const passhashify = require('../utilities/passhashify')
+
+const SELF_SELECT =   "-auth -preferred_position";
+const COACH_SELECT =  "-auth";
+const SAFE_SELECT =   "-auth -weight -preferred_side -preferred_position";
```
k3ntako: Seems like these aren't being used anywhere. But, I'm also not sure what these are for.

client/pages/Event/index.js
```diff
@@ -39,111 +41,67 @@ export default class EventPage extends Component {
...
         <RippleButton icon='edit' floating primary />
       </Link>
+    } else {
+      let response = event.findResponseForPerson(me);
+      return <div className={styles.eventActions}>
+        <RippleButton
+          icon='clear'
+          floating primary={response === false}
+          className={styles.action}
+          onClick={() => this.toggleAttending(false)}
```
k3ntako: I'm having an issue where I have to click on the center of these attendance buttons. Outer doesn't trigger the `onClick` event handler.

client/pages/Boat/SeatPicker.js
```diff
@@ -1,5 +1,5 @@
 import React, {Component} from 'react';
-import Autocomplete from 'react-toolbox/lib/autocomplete/Autocomplete';
+import Autocomplete from 'react-toolbox/lib/autocomplete';
```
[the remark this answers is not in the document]
mienaikoe: I'll revert this - The SeatPicker was just throwing errors and I needed someone in a seat, so this was a temporary fix that I forgot to take out at the end.

server/models/Person.js
mienaikoe: I mean, if we were going to make a special case for pacers, why not make this field D, S, and 1-10

I wanted to keep letters to match the seat. It's harder to mess up capitalization when it's one letter. If we were using typescript, different story.

client/pages/Event/index.js
mienaikoe: I don't have this problem. Can you be more specific?

server/models/Person.js
```diff
@@ -3,11 +3,14 @@ const mongoose = require('mongoose')
...
-  "L","R"
+  "L","R",null
 ];
 
 const PREFERRED_POSITIONS = [
-  "P","E","R","D","S"
+  "D",
+  "1","2","3","4","5",
+  "6","7","8","9","10",
+  "S"
```
k3ntako: I think you need null here.

server/models/Person.js
k3ntako: Hey I feel bad bring this up, but I'm wondering what we should include for preferred_position. 

I really liked having P, E, and R, because people are generally pretty fluid within each section. I bought up row 1 because I felt like that might be the exception. 

What do you think?

PR #23, opened 2019-01-30
Boat page

k3ntako:


k3ntako:
Issue: Scroll bar shows up on non-webkit browsers. 

Can't use this trick to scroll bar because the border would be pushed down 17px.
https://stackoverflow.com/questions/16670931/hide-scroll-bar-but-while-still-being-able-to-scroll

client/pages/Boat/Seats.js
```diff
@@ -38,46 +48,100 @@ export default class Seats extends Component {
...
-          <div key={`rowNum-${idx}`} className={styles.rowNum}>{ (idx + 1)/2 }</div>
+          <div key={`rowNum-${idx}`} className={styles.rowNum}>
+            <div><span>{ (idx + 1)/2 }</span></div>
+          </div>
+        );
+      }
+
+      if(idx === 6 || idx === 14){
+        const aboveText = idx === 6 ? "Pacers" : "Engine";
+        const belowText = idx === 6 ? "Engine" : "Rocket";
```
mienaikoe: I think this could be "rockets" plural instead of "rocket"

Same for Engines

client/pages/Boat/index.css
```diff
@@ -1,39 +1,105 @@
 .seatsContainer{
+  font-family: 'Oswald', sans-serif;
   width: calc(78vw + 30px);
```
mienaikoe: avoid calc!

Also, avoid vw. Whatever this is calculating can probably be done by changing box-sizing or using flex.

client/pages/Boat/index.css
```diff
@@ -1,39 +1,105 @@
...
 }
 
 .seat:first-of-type, .seat:last-of-type{
+  margin: 0;
   grid-column: 2 / span 3;
+  text-align: center;
+}
+
+.coachModeInput{
+  font-family: 'Oswald', sans-serif;
```
mienaikoe: I don't think the inputs should be in Oswald. So far they've all been in Comfortaa and they should stay that way. Oswald should only be for display.

client/pages/Boat/index.js
```diff
@@ -18,92 +19,127 @@ export default class BoatPage extends Component {
...
 
-  renderSeats( meState ){
+  windowResizeHandler(){
+    let timer;
+    let self = this;
+
+    return function(event){
+      if(timer) clearTimeout(timer);
+      timer = setTimeout((() => {
+        const isMobile = window.innerWidth <= 600;
```
mienaikoe: Oh ugh. You really couldn't do this with css?

mienaikoe: You can render both elements and then only show one of them with `display: none;` based on the max-width. Maybe more memory-hungry, but you really want to avoid this wherever you can. Css will always give you better rendering performance because it's running partially in the gpu.

client/pages/Boat/index.js
```diff
@@ -18,92 +19,127 @@ export default class BoatPage extends Component {
...
+
+    return<Tab key={boat.name} label={boat.name} className={styles.boatTab}>
+      <div className={styles.seatsContainer}>
+        <Seats
+          isCoach={isCoach}
+          seatAssignments={seatAssignments}
+          team={team}
+          boat={boat}
+          boats={boats}
+          saveSeat={this.saveSeat}
```
mienaikoe: why can't this be further down in the hierarchy? The Seats component should handle things having to do with seats unless this component needs to tell other components about changes regarding seats.

client/pages/Boat/SeatSuggestions.js
```diff
@@ -0,0 +1,53 @@
...
+
+
+export default (props) => {
+
+  function sortPeople(){
+    let team = props.team;
+    let suggestedPeople;
+
+    if(props.search){
+      suggestedPeople = matchSorter(team, props.search, {keys: ['name']})
```
mienaikoe: This has `left-pad` written all over it. You should really avoid importing libraries for small, individual actions that you could easily do yourself.

check out `Array.prototype.sort`. 

And read some articles about react and left-pad...

client/pages/Boat/BoatListFunctions.js
```diff
@@ -0,0 +1,49 @@
...
+  Object.keys(allBoatsAssignments).forEach(key => {
+    if(allBoatsAssignments[key].people.includes(person.id)){
+      boatsList.push(allBoatsAssignments[key].name);
+    }
+  });
+
+  let boatsName = [];
+  const boatsLength = boatsList.length;
+  if(boatsLength > 3){
+    boatsName = boatShortenedList(boatsList, boatsLength);
```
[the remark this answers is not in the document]
mienaikoe: Nah I would just not worry about it right now. Focus on getting this out the door then we can get fancy.

client/pages/Boat/SeatPicker.js
```diff
@@ -1,60 +1,49 @@
...
+    let allBoatsAssignments = this.props.allBoatsAssignments;
+
+    let boatsName = BoatFunctions.generateBoatList(person, allBoatsAssignments);
 
     return <div>
-      <div>Weight: { weight || "N/A" }</div>
-      <div>Prefered Side: { prefered_side || "N/A" }</div>
-      <div>Preferred Position: { prefered_position || "N/A" }</div>
+      <div><strong>Weight:</strong> { weight || "N/A" }</div>
+      <div><strong>Prefered Side:</strong> { prefered_side || "N/A" }</div>
```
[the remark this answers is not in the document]
mienaikoe: happens. I've misspelled lots of things in my career.

client/pages/Boat/SeatSuggestions.js
mienaikoe:
```
match = match.toLowerCase();
name.toLowerCase().includes(match)
```
You taught me that one!

client/pages/Boat/index.js
mienaikoe: Yea I guess it's not really a performance thing. It's more of a code-methods thing. If you do one thing one way, you should follow that pattern everywhere else in the project so you don't end up creating a giant mess of spaghetti that can't be debugged.

client/pages/Boat/index.js
mienaikoe: You should only need to re-render the seats component. There's nothing else in this component that would change.

client/pages/Boat/BoatListFunctions.js
```diff
@@ -0,0 +1,49 @@
+function generateBoatList(person, allBoatsAssignments){
+  let boatsList = [];
+  Object.keys(allBoatsAssignments).forEach(key => {
```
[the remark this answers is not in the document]
k3ntako: I tried this, but it tells me that `allBoatsAssignments.forEach is not a function`.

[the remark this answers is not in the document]
mienaikoe: Oh I misunderstood what allBoatsAssignments was caus it was quite buried. It's an object? I guess I don't get what the point is to make it an object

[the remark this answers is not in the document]
k3ntako: Yeah it's much more complicated than it had to be. I'm rewriting it and it looks much simpler.

client/pages/Boat/SeatSuggestions.js
```diff
@@ -0,0 +1,53 @@
...
+    let boatsName = BoatFunctions.generateBoatList(person, allBoatsAssignments);
+
+    return <div key={person.id} className={styles.suggestion} onMouseDown={() => assignSeat(person)}>
+      <div className={styles.name}><strong>{person.name}</strong></div>
+      <div className={styles.boatsList}>{boatsName}</div>
+    </div>
+  });
+
+  return <div className={styles.suggestionsContainer}>
+    <div className={styles.suggestions}>
```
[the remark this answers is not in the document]
k3ntako: There might be a way. The reason I made it two was so I can change the size of the `.suggestions` without changing the size of the whole row (`.suggestionsContainer`). This is what I ended up with when I had one div.

screen shot 2019-01-31 at 1 19 31 pm


```
.suggestionsContainer{
  font-family: 'Oswald';
  font-size: 1.3rem;
  position: absolute;
  width: 100%;
  z-index: 1;
}

.suggestions{
  position: relative;
  top: -20px;
  background-color: white;
}
```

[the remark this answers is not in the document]
mienaikoe: Ohhh caus you need something to base the absolute positioned dropdown from.

Btw, `.suggestions` should probably be the thing that's positioned absolute, within the relatively positioned container.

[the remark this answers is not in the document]
k3ntako: What would I name the class that is currently `.suggestionsContainer`?

client/pages/Boat/BoatListFunctions.js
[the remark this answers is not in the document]
k3ntako: I left this open because I'm not sure I like the way I approached it.

client/pages/Boat/index.js
k3ntako: I think `saveSeat()` does have to be in the top. 

Basically, `index.js` and `Seats.js` are on the same level. It was originally one file but I broke it up into two when I thought `index.js` became too large. Is there a better way to deal with bigger files with lots of functions?

client/pages/Boat/SeatSuggestions.js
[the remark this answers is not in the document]
mienaikoe: no - keep the html/react as it is. just change the css

client/pages/Boat/index.js
mienaikoe: Components with lots of functions can probably be broken down into multiple components. That's not always the case, but 9 times out of 10, it is.

usually how I approach it is to figure out where there are divisions in the use of props and state. If there's a set of state variables that are only used by a few functions, then that's a good sign that a component can be broken out there.

mienaikoe: To clarify why I suggested to move it down:

The further down you move a `forceUpdate` or `setState` call, the more performant your code will be. This also typically reduces the occurrence of monolithic components that try to handle everything and don't defer tasks to children.

k3ntako: Thanks for the feedback!

PR #24, opened 2019-02-03
Adding Breadcrumbs and tab path params

mienaikoe:
- Minor Redesign Pass for Boat Page
- Adding Breadcrumbs to Boat Page
- Adding Path Params to event page and boat page


client/pages/Event/index.js
```diff
@@ -19,13 +19,25 @@ import Heading from '../../components/Heading';
...
+]
+
+
+class EventPage extends Component {
 
   constructor(props){
     super(props);
     this.state = {
-      tab: 0
+      tab: TAB_NAMES[document.location.hash] || 0
```
k3ntako: Is `document.location.hash` supposed to be `props.match.params.tab`?

client/pages/Event/index.js
```diff
@@ -123,3 +146,6 @@ export default class EventPage extends Component {
     </div>
   }
 }
+
+
+export default withRouter(EventPage)
```
k3ntako: What is `withRouter()` for? It seems to work without it. I think you still have access to `props.match.params` without it.

client/pages/Boat/index.css
```diff
@@ -106,8 +107,12 @@
 }
 
 @media screen and (max-width: 600px){
+  .seatsContainer{
+    width: 100%;
```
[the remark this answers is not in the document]
mienaikoe: The extra white-space was annoying me. Not such a problem on mobile.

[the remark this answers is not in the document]
mienaikoe: I've taken a look at 600px, and it looks fine in white space. It was just the full width of 960 that was bothering me.

client/pages/Event/index.js
mienaikoe: yes

client/pages/Event/index.js
mienaikoe: Uhh that's not supposed to happen. `withRouter` is supposed to supply router props to your component, like match, history, etc.

https://github.com/ReactTraining/react-router/issues/5345

apparently Route children directly get it for free. But... it still bothers me that we rely on knowing what the parent is to render correctly. In React, children aren't supposed to reach up into their parents - they should be ambivalent to whatever goes on above them.

k3ntako: This is so strange. Let's keep it!

PR #26, opened 2019-02-04
Change and Reset Password

k3ntako:


client/pages/Me/ChangePassword.js
```diff
@@ -0,0 +1,80 @@
...
+    let email = this.props.me.email;
+
+    Person.updatePassword( oldPass, newPass ).then( () => {
+      this.props.issueSnackbar("Your password was saved");
+    }, error => {
+      const incorrect = "The old password you have entered is incorrect";
+      const other = "Could not save your password: " + error.message
+      let message = error.message === "Password Did Not Match" ? incorrect : other;
+
+      this.props.issueSnackbar(message);
```
k3ntako: I think this error message should be more than a snackBar. I need to change this

client/pages/Auth/Login/LoginPage.js
```diff
@@ -78,8 +101,10 @@ class LoginPage extends React.Component {
               disabled={!this.valid()} />
           </div>
           <div className={styles.messages}>
+            { this.renderSuccess() }
             { this.renderError() }
```
mienaikoe: These could be snackbars from react-toolbox, set above `Routes`. You don't really have a reason to keep them around indefinitely.

Before, when these were rendering errors, the errors pertained to the login process. Messages from the Reset Password page shouldn't be handled by the Login Page

client/pages/Auth/RequestPassReset/Message.js
```diff
@@ -0,0 +1,33 @@
...
+
+  constructor(props){
+    super(props);
+    this.state = {};
+  }
+
+  renderSuccess(){
+    return <div className={styles.requestMessage}>
+      <h2>Email Sent!</h2>
+      <p>An email was sent to {this.props.email} with instructions on reseting your password. Please reset your password within 24 hours.</p>
```
mienaikoe: spelling: `reseting` should be `resetting`

client/pages/Auth/ResetPassword/index.js
```diff
@@ -0,0 +1,96 @@
...
+      error: null,
+      email: null,
+      token: null
+    };
+  }
+
+  componentDidMount(){
+    let queryString = this.props.location.search;
+
+    if(queryString.slice(0,7) !== "?token="){
```
mienaikoe: Could just use `startsWith`

client/pages/Me/ChangePassword.js
mienaikoe: Yea. Errors native to a page should be more persistent than a snackbar. Try the same thing we did for login/join, where we used red text or a red box with red text next to the button to explain what happened.

client/pages/Me/MePage.js
```diff
@@ -47,18 +53,29 @@ export default class MePage extends Component {
...
-            value={this.state.first_name}
-            onChange={this.handleChange} />
-          <Input type='text' label='Last Name'
-            value={this.state.last_name}
-            onChange={this.handleChange} />
-          <Input type='number' label='Weight (lbs)'
-            value={this.state.weight}
-            onChange={this.handleWeightChange} />
+        <div className={styles.form}>
+          <div>
```
mienaikoe: Why not put these both in `Card`s instead of plain divs? Also, these are separate forms with separate actions, so they should be separated in the UI. Two cards would be better than styling the entire page to have a white background.

client/pages/Me/index.css
```diff
@@ -1,4 +1,13 @@
+.form{
+  padding: 2rem 2rem;
+  background-color: white;
+  margin: 2rem auto;
+  box-shadow: 0 6px 20px rgba(0,0,0,0.3);
```
mienaikoe: Use a `<Card>`

server/controllers/PersonController.js
```diff
@@ -186,6 +187,114 @@ const PersonController = {
...
+      throw new ValidationError("You're already logged in")
+    }
+
+    const email = req.body.email.toLowerCase();
+    const resetToken = generateResetToken();
+
+    return Person.
+      findOne({ email }).
+      then ( (person) => {
+        const tomorrow = new Date().getTime() + (3600 * 24 * 1000);
```
mienaikoe: `3600 * 24 * 1000` could be a constant. This is commonly used to prevent devs from not understanding what a series of numbers means.

server/controllers/PersonController.js
```diff
@@ -186,6 +187,114 @@ const PersonController = {
       })
   },
 
+  updatePassword: function(req, res, next){
+    return Person.
+      findOne({ _id: req.session.person_id }).
+      then((person) => {
+        if( !person ){
```
mienaikoe: if you use `requireAuth` in `routes` you shouldn't need to do this check.

server/routes.js
```diff
@@ -23,6 +23,9 @@ function requireAuth(req, res, next){
 router.post(  '/join',    PersonController.join)
 router.post(  '/login',   PersonController.login)
 router.post(  '/logout',  PersonController.logout)
+router.post(  '/update_password',  PersonController.updatePassword)
```
mienaikoe: use `requireAuth` for this route

server/utilities/resetPassword.js
```diff
@@ -0,0 +1,40 @@
...
+sendPasswordEmail = (person, token) => {
+  let { email, first_name } = person;
+
+  const link = `${process.env.URL}/reset-password?token=${token}&email=${email}`;
+
+  const message = {
+    from: `${process.env.GMAIL}@gmail.com`,
+    to: email,
+    subject: "Mighty Speed Krush: Reset Password",
+    text: `Hello ${first_name},\n\n${messageParts.intro}\n\n${messageParts.reset} ${link}\n\n${messageParts.timeLimit}\n\n${messageParts.signature}`,
```
mienaikoe: newlines are conserved in backtick string templates. You can supply them directly instead of typing \n\n everywhere.

What might be even better is to put the body in its own js function so you can format it however you like without making this function look messy.

test/server/PersonControllerTest.js
```diff
@@ -119,7 +119,7 @@ describe('PersonController', function() {
           assert(req.session.team_id)
           assert(response);
           assert(response._id);
-          assert.equal(response.email, "TestUser@mskdragons.com");
+          assert.equal(response.email, "TestUser@mskdragons.com".toLowerCase());
```
mienaikoe: you don't need to call `toLowerCase` on a string you have control over. Just make the string itself be lowercase.

test/server/PersonControllerTest.js
```diff
@@ -119,7 +119,7 @@ describe('PersonController', function() {
...
           assert.equal(response.first_name, 'Test');
           assert.equal(response.last_name, 'User');
           assert.equal(response.weight, null);
```
mienaikoe: You should write tests for the reset password endpoints.

k3ntako: I haven't been able to write tests for `Person.resetPassword()`, because it takes the reset token as an argument. I hashed the token, so I don't have access to the token without the email.

mienaikoe: You can test everything besides the email send. Just make sure the token is returned from some function, store it, and then verify that using that token works when validating. You can call two controller functions in a test. It doesn't have to be one.

test/server/BoatControllerTest.js
```diff
@@ -57,7 +57,6 @@ describe('BoatController', function() {
 
   describe('#find()', function() {
     before(function(done){
-      console.log(TestHelper.session().person_id);
```
k3ntako: You don't need this, right?

test/server/TeamControllerTest.js
```diff
@@ -27,8 +27,6 @@ describe('TeamController', function() {
         json: function(responseDoc){
           let response = JSON.parse(JSON.stringify(responseDoc));
           expect(response._id).to.equal(test_session.team_id.toString());
-          console.log(response.members);
-          console.log(test_session);
```
k3ntako: You don't need this, right?

PR #28, opened 2019-02-07
Fixes to various things witnessed on Feb6

mienaikoe:


migrations/0_CreateMSK.js
```diff
@@ -12,14 +12,14 @@ var james;
 
 
 function createJames(){
-	return Person.findOne({ email: "james.lozada@mskcc.org" }).
+	return Person.findOne({ email: "james.lozada@gmail.com" }).
```
k3ntako: is this his real email? Don't think he wants that out there?

PR #29, opened 2019-02-07
Event Note

k3ntako:
Things I would especially like your feedback on:
- Input for event notes is currently the full width of the form.
- It seemed weird to have two notes (event and location) so I hid the location one with an icon to reveal it. 
- Made the nav-bar not selectable (CSS: `user-select: none;`). Wondering how you feel about that. This is kinda random.


client/pages/Event/EventDetails.js
```diff
@@ -1,44 +1,91 @@
...
+
+  toggleLocNote(){
+    this.setState({locNoteVisible: !this.state.locNoteVisible});
+  }
+
+  renderLocNote(locNote){
+    let icon = 'keyboard_arrow_down';
+    let inside;
+    if(this.state.locNoteVisible){
+      icon = 'keyboard_arrow_up';
```
mienaikoe: Honestly, I didn't even see this icon and would have missed it if I were a new user. I have a feeling Coach will just need to put the location details in once or twice for practice. For races, he'll put it in an email, or would be part of the event note. Maybe we can just remove the note field on location?

client/components/Nav/index.css
```diff
@@ -9,6 +9,7 @@
   flex-direction: row;
   align-items: center;
   justify-content: space-between;
+  user-select: none;
```
mienaikoe: Ah. Had to try it out to understand why. Yea this is good.

client/pages/EventForm/EventForm.js
```diff
@@ -273,23 +285,20 @@ export default class EventForm extends Component {
...
+              location={location} />
           </div>
           <div className={styles.actions}>
             <Button
               raised
               primary
               onClick={this.submitEventForm}
-              label="Submit" disabled={!!this.validateForm().length}
-            />
+              label="Submit" disabled={!!this.validateForm().length} />
```
mienaikoe: It's perfectly fine to just say `this.validateForm().length` to evaluate a truthy statement.

client/pages/Event/EventDetails.js
k3ntako: I definitely had the same thought, but I feel like if we are gonna build an app we should put all the info here. I think people lose emails, and it would be nice to know that you can count on this app to have all the info. 

I'm wondering if it would be a good idea to have a page for locations. A simple page with name, address, note and map.

mienaikoe: I get what you're saying in principle, but I just don't know if location note is really that useful. If "All the info" is the goal, we'd overload users with too much information.

k3ntako: I was thinking it could be used for keeping directions and stuff like that. Do you think having a page for each location would be overkill? Instead of the name of the location going to Google Maps, it can go to that page.

mienaikoe: I think have a separate page would be overkill, yea. Google Maps will always do things 100x better than we can.

client/pages/EventForm/EventForm.js
mienaikoe: eh good enough

client/pages/Event/EventDetails.js
k3ntako: Ahh good point. okay we can see how people react to it

PR #30, opened 2019-02-08
Events Page Fixes + Boat Page Center of Mass

mienaikoe:
- Empty State
- Sorting of Events
- Filtering out Events in the past


client/models/Boat.js
```diff
@@ -2,6 +2,34 @@ import Model from './Model';
...
+	"8L":[-25,-170],
+	"8R":[25,-170],
+	"9L":[-22,-248],
+	"9R":[22,-248],
+	"10L":[-19,-326],
+	"10R":[19,-326],
+	"S":[0,-435]
+}
+
+const BOAT_MASS = 280; // kg
```
k3ntako: I see that you're using kg here, but where are you converting the pounds to kg?

client/models/Boat.js
```diff
@@ -2,6 +2,34 @@ import Model from './Model';
 import Person from './Person';
 import Fetcher from './Fetcher';
 
+// assuming distances of 78 cm per seat, and an average forward lean of 25 cm
+const SEAT_DISTANCES = {
+  "D":[0,485],
+	"1L":[-19,376],
+	"1R":[19,376],
```
k3ntako: indentation is a little weird

client/models/Person.js
```diff
@@ -104,6 +104,14 @@ class Person extends Model {
     return `${this.first_name[0].toUpperCase()}${this.last_name[0].toUpperCase()}`;
   }
 
+  set weight( pounds ){
+    this._weight = pounds;
```
k3ntako: I'm a little confused about the `set weight()` and `get weight()`. You aren't using it and the function doesn't update the database.

client/pages/Boat/BoatPanel.js
```diff
@@ -70,27 +64,28 @@ export default class BoatPanel extends Component {
...
     } else {
       followup = Promise.resolve();
     }
 
-    this.state.boat.seat(person, seatName).then(followup).then( () => {
-      this.forceUpdate();
-    })
+    this.props.boat.seat(person, seatName).
+      then(followup).
+      then(this.props.onSeat);
```
k3ntako: Do you not need `.catch` blocks when you have promises?

client/pages/Boat/BoatWeight.js
```diff
@@ -0,0 +1,65 @@
...
+
+
+export default class BoatWeight extends Component {
+
+  render(){
+    let centerOfMass = this.props.boat.centerOfMass()
+
+    let graphedCenterOfMass = {
+      x: Math.max( Math.min(centerOfMass.x, 80), -80),
+      y: Math.max( Math.min(centerOfMass.y, 80), -80),
```
k3ntako: where does `80` and `-80` come from?

client/models/Boat.js
mienaikoe: Ah yea. I need to do that.

client/models/Person.js
mienaikoe: this is a setter and getter. You can call them like a field, and the field will actually call the function.

I'm using the getter to ensure that a weight comes back even if it hasn't been set in the API. This is because when calculating center of mass, having one person missing would make a big difference

client/pages/Boat/BoatWeight.js
mienaikoe: These are thresholds beyond which I don't want to render and end up cutting off the text.

client/pages/Boat/BoatPanel.js
mienaikoe: hmm I hadn't thought about error catching.

Now that I think about it, I'm not sure what to do. Maybe a snackbar?

k3ntako: Yeah i think that would be good

mienaikoe:
> I noticed that design issues with the center of mass graph. If you minimize the page, it gets cut off around 650px.
> 
> I noticed that the dividers between the sections on the boat is now Comfortaa. I did like it better if as Oswald. in my mind I thought of Oswald as titles. Doesn't look like you purposely changed it?
> 
> When the graph is at (0,0) the numbers overlap each other. Is there an easy fix? If not, I don't think its a big deal.

Could you put these as line items in the code next time? It's hard to follow each one separately when they're all thrown in a block of text at the top.

client/pages/Boat/index.css
```diff
@@ -67,10 +67,11 @@
 }
 
 .seat{
+  flex: 1;
   font-family: 'Oswald', sans-serif;
   position: relative;
   font-size: 1.5rem;
-  width: 240px;
+  /* max-width: 240px; */
```
k3ntako: Can we remove this?

Also, I noticed that removing this is causes an issue with the dropdowns for the steer and drummer. The suggestions end up being too wide and cut off. In `SeatSuggestions.css` you can remove the following to fix it:

```
.suggestions{
  width: 150%;
...
}
```
```
.suggestions.center{
  left: -25%;
}
```

client/pages/Boat/index.css
```diff
@@ -123,7 +166,15 @@
...
 }
 
-@media screen and (max-width: 600px){
+@media screen and (max-width: 720px){
+  .boatTab{
+    display: block;
+  }
+
+  .centerOfMassContainer{
+    width: 100%;
```
k3ntako: This isn't a big issue, so feel free to ignore it. I noticed that when the screen size goes below 720px on a computer the graph is too tall to fit on the screen. But it looks fine on a phone (via Chrome's inspector and my iPhone). 

Probably not a big issue because most people won't make their window so small.

screen shot 2019-02-09 at 10 46 44 am

client/pages/Boat/index.css
mienaikoe: done

client/pages/Boat/index.css
mienaikoe: I'll just set the breakpoint a bit higher then, caus I don't want either of those to get smaller.

mienaikoe: done

PR #31, opened 2019-02-09
Eliminate alerts

k3ntako:
I think you won't like some of these. I'm open to suggestions.


client/pages/Event/EventBoats.js
```diff
@@ -40,7 +40,7 @@ export default class EventBoats extends Component {
         this.props.history.push(`/events/${eventId}/boats/${boatId}`);
       });
     }else{
-      alert("Boat name cannot be blank.")
+      this.props.issueSnackbar("Boat name cannot be blank.");
```
mienaikoe: This whole check should be upstream. You don't need an alert. Put a boolean on the dialog submit button for `disabled` and make it disabled if the title is all whitespace.

client/pages/EventForm/EventForm.js
```diff
@@ -187,14 +187,18 @@ export default class EventForm extends Component {
...
+      this.setState({error: true})
     });
   }
 
 
   render(){
     const pageTitle = this.props.match.path.includes("/edit") ? "Edit Event" : "Add New Event";
+    console.log(this.state.error);
+    if(this.state.error){
+      const currentPath = this.props.history.location.pathname;
```
mienaikoe: This just has them reload the page? Maybe we should suggest that they go back to the event page if they're editing, or the upcoming events page if they're creating.

client/pages/EventForm/EventMapContainer.js
```diff
@@ -56,25 +57,33 @@ export default class EventMapContainer extends Component {
...
+    const lat = nextCenter.lat();
+    const lng = nextCenter.lng();
+    const address = place.formatted_address;
+    const googleId = place.place_id;
+
+    this.props.updateLocation(place.name, address, lat, lng, googleId);
+  }
+
+  noResult(){
+    this.setState({error: true}, () => {
```
mienaikoe: I don't understand the purpose of this function. could you add some comments?

client/pages/EventForm/FormError.js
```diff
@@ -0,0 +1,24 @@
...
+
+
+export default function(props) {
+  console.log(props);
+
+
+  return <div className='page'>
+    <Heading title={props.pageTitle} />
+    <section>
+      <Card className={styles.formError}>
```
mienaikoe: I've been doing it where empty states and error states are not inside of cards, but are rendered in centered large text directly on the background. This gives a clear design expectation that cards are reserved for actual content.

See the Events page empty state for an example.

client/pages/EventForm/FormError.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
```
mienaikoe: Nice thinking, making this reusable

client/pages/EventForm/index.js
```diff
@@ -9,20 +10,24 @@ export default class EventFormContainer extends Component {
   constructor(props){
     super(props);
     this.state = {}
+    this.renderForm = this.renderForm.bind(this);
   }
 
-  renderForm(meState){
+  renderForm(meState, issueSnackbar){
```
mienaikoe: you don't really need this function. It's only called from one place and the render function isn't that big.

client/pages/EventForm/EventMapContainer.js
k3ntako: THis is one I thought you wouldn't like. It flashes translucent notification saying "No Result" if Google Maps returns no result. Image below.

Now, I'm thinking that the user can figure it out without any type of notification?

screen shot 2019-02-10 at 12 18 31 pm

mienaikoe: Hmm if you gray out the map, the user can't change their input preferences though.

Yea I'm not a fan of flashes. But here, a snackbar also doesn't quite work.

I think for now, coach knows how to use google maps autocomplete, so we can just leave it to do nothing when there are no results.

mienaikoe:
There's a merge conflict. Fix it and i'll merge.

PR #34, opened 2019-02-12
Delete Event and Boat

k3ntako:
Ability for coach to delete event and boat.


mienaikoe:
🌟  I'll check a bit later

client/pages/Boat/BoatTab.js
```diff
@@ -32,9 +62,17 @@ export default class BoatTab extends Component {
...
+    if( this.props.isCoach ){
+      deleteButton = this.renderDelete();
+    }
+
     return <div className={styles.boatTab}>
       <BoatPanel {...this.props} boat={this.state.boat} onSeat={this.forceUpdate.bind(this)} />
-      <BoatWeight boat={this.state.boat} />
+      <div>
+        <BoatWeight boat={this.state.boat} />
+        {deleteButton}
```
mienaikoe: A full width button below the weight doesn't seem like the right place to put it. We may fill that sidebar with more things, and the button would be in the way.  Why not fix it to the bottom right corner of the screen? I'm thinking of making an edit boat page as well, so handling edit/delete the same way for events and boats would be nice because it's an expected way for things to work.

client/pages/EventForm/index.css
```diff
@@ -18,13 +18,30 @@
...
   margin: 2rem 1rem 1rem;
 }
 
+.deleteButton{
+  background-color: var(--palette-grey-300) !important;
+  color: var(--palette-red-500) !important;
+}
+
+.deleteButton:hover{
+  background-color: var(--palette-red-500) !important;
```
mienaikoe: The coloring on these delete buttons are different that the ones you made in `/components`. Why not use the same component, or at least the same coloring.

client/components/DeleteConfirmation/index.js
```diff
@@ -0,0 +1,39 @@
...
+export default class EventAttendance extends Component {
+  constructor(props){
+    super(props)
+    this.state = {};
+    this.deleteObject = this.deleteObject.bind(this);
+  }
+
+  deleteObject(){
+    this.props.object.del().then(() => {
+      this.props.history.push(this.props.redirect || '/');
```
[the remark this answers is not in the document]
k3ntako: So the idea here is that it will go to `this.props.redirect` and if that's undefined it defaults to `/`.

client/pages/Boat/BoatTab.js
k3ntako: So you're suggesting that I change the delete button to a circular floating button? 

I was thinking about that first but here are the reasons why I chose not to:
1. I thought it was good to change the style so it's not mistaken for something else. 
2. The floating button felt very prominent for something that won't be used often.
3. The floating button on mobile (in the bottom right) might get in the way of boat and the graph. Haven't tried it out.

This is not really a reason because we have to fix this on other pages too, but here's another issue with the floating buttons:
screen shot 2019-02-12 at 10 00 35 pm

client/pages/EventForm/index.css
k3ntako: On the Event and Boat pages I wanted it to be a little more muted, because it's not a button that will be clicked often. But, on the confirmation I wanted it to catch the eye.

client/components/DeleteConfirmation/index.js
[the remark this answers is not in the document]
mienaikoe: ah I misread, sorry. Have had a rough day.

client/pages/Boat/BoatTab.js
mienaikoe: I suggested making it a circle button, and adding it to the bottom right corner of the screen.

mienaikoe: And made another request to put it in an ellipsis button that opens up to many more buttons. It's ok if you don't want to do that part right now, but at least bottom right would be a good spot for now.

client/pages/EventForm/index.css
mienaikoe: Why did you want it to catch the eye?

client/pages/Boat/BoatTab.js
mienaikoe: I'll do it in another commit

k3ntako: I changed the delete to a floating button. I also added an ellipsis button that opens to a menu. I feel like the design could be improved

client/pages/EventForm/EventForm.js
```diff
@@ -192,7 +192,8 @@ export default class EventForm extends Component {
 
 
   render(){
-    const pageTitle = this.props.match.path.includes("/edit") ? "Edit Event" : "Add New Event";
+    let pageTitle = "Add New Event";
```
mienaikoe: Edit Event Title should be here

client/pages/Boat/index.js
```diff
@@ -76,6 +91,43 @@ class BoatPage extends Component {
...
+
+  renderDelete(meState){
+    if(!meState.me.isCoach() || !this.state.event){
+      return null;
+    }
+
+    let link = this.props.history.location.pathname;
+    link = link.slice(0, link.indexOf("/boat"));
+
+    let deleteBoat = () => {
```
mienaikoe: define this one layer above.

k3ntako:
What it looks like:

screen shot 2019-02-15 at 7 07 56 pm

mienaikoe:
Spacing is a bit odd, but we can fix it in post.

PR #35, opened 2019-02-12
Bug Fixes 

k3ntako:
- An unauthorized user would receive `null` for `me`. A null check now sends them to the login page.
- Add event button was missing from events page.


client/pages/Team/TeamPage.js
```diff
@@ -91,9 +92,22 @@ class TeamPage extends Component {
...
+        <RippleButton icon='add' floating primary />
+      </Link>
+    }
+    return null
+  }
+
   render(){
     let me = this.props.me;
 
+    if(!me){
```
mienaikoe: This is already handled in MeContext. You don't need to put it on every page.

k3ntako: I think the check on `MeContext` isn't working. This is my thought on it:

When I load 'localhost:3000/' without being logged in, here is the order of methods called:
1. Render
2. componentDidMount 
3. onHasMeNot
4. setState(meChecked: true)
5. Render
6. afterMeRoute

At step 5, `this.state` is `{ me: null, meChecked: true }`. This returns the provider with `me` as `null`,  which causes errors on `TeamPage` (e.g. `me.team.name`).

I have an idea for how to fix this.

client/contexts/MeContext.js
```diff
@@ -68,6 +68,23 @@ class MeProviderComponent extends React.Component {
...
+    ]
+
+    const includesPath = paths.filter(path => nextProps.location.pathname.includes(path)).length;
+
+    if(!includesPath && nextState.meChecked && !nextState.me){
+      this.afterMeRoute()
+      return false;
+    }
+    return true
+  }
```
k3ntako: this is the proposed fix

client/pages/Team/TeamPage.js
mienaikoe: I wouldn't write any of the pages with the assumption that me is not null. There may be a render before MeContext handles the route.

client/contexts/MeContext.js
mienaikoe: I like it.

client/pages/Team/TeamPage.js
k3ntako: I think I'm a bit confused by what you mean. Are you saying that I should have a null check in TeamPages (and other pages with me?).

mienaikoe: Yea that's what I was saying. While the use case of these pages requires a logged-in user, there's inevitably going to be something that causes them to get into a state of a logged-out user seeing something. Best course of action would be to render null.

But your change to MeContext is the ideal solution to the underlying problem so I'm glad you did it.

PR #37, opened 2019-02-17
Change Boat Name and Visibility

k3ntako:
This was built on top the other PR that's still open.


client/pages/Boat/index.js
```diff
@@ -76,8 +106,74 @@ class BoatPage extends Component {
...
+  }
+
+  renderMenu(meState){
+    if(!meState.me.isCoach() || !this.state.event){
+      return null;
+    }
+
+    let boat = this.state.event.boats[this.state.index];
+    let icon = boat.isVisible ? "lock" : "lock_open";
+    let caption = boat.isVisible ? "Hide Boat" : "Show Boat";
```
mienaikoe: I'm worried that the wording is too vague here (who are we hiding it from?). Maybe "Show to Team" and "Hide from Team". Also open to suggestions.

client/pages/Boat/index.js
```diff
@@ -95,9 +191,35 @@ class BoatPage extends Component {
...
 
+    let menu = <MeConsumer>
+      {this.renderMenu}
+    </MeConsumer>
+
+    let link = this.props.history.location.pathname;
+    link = link.slice(0, link.indexOf("/boat"));
+
+    if(this.state.event){
+      boat = this.state.event.boats[this.state.index];
```
mienaikoe: This check is done already at line 178. Best to combine checks unless some logic between changes the circumstances in which the checks occur. (Not the case here)

client/pages/Boat/index.js
```diff
@@ -95,9 +191,35 @@ class BoatPage extends Component {
       </Link>
     ];
 
+    let menu = <MeConsumer>
```
mienaikoe: Why not just wrap this whole component in MeConsumer?

Or make the Menu its own component and have the MeConsumer there. <- Actually I like this better.

client/pages/Event/EventBoats.js
```diff
@@ -45,19 +45,25 @@ export default class EventBoats extends Component {
...
+  makeListBoats(meState, boatsArr){
+    const isCoach = meState.me.isCoach();
     const eventId = this.props.event.id;
-    let boats = contentsArr.map(content => {
-      return <div className={styles.eventPageList} key={content._id}>
-        <Link to={`/events/${eventId}/boats/${content.id}`}>
-          <BoatCard boat={content} />
+
+    let boats = boatsArr.map(boat => {
+      if(!isCoach && !boat.isVisible){
```
mienaikoe: This is used often enough, it might be nice to add a function to the boat model that just says `boat.isVisibleTo(me)` and it just returns the result of this check.

client/pages/Boat/index.css
```diff
@@ -131,7 +131,7 @@
 }
 
 .tabCenter{
-  margin-top: -47px;
+  margin-top: -48px;
```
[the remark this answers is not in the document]
k3ntako: The button was shorter than the height of the background, so there was a gap between the ripple and the header. You can see it above "Upcoming Events". Unless that was on purpose, I thought it was better without the gap.

screen shot 2019-02-21 at 12 40 17 pm

client/pages/Boat/index.js
k3ntako: This menu should only be displayed if the user is a coach. But, that wouldn't necessarily the case for all menus? If I do the check for coach in the menu, only coach could see any of the menus.

client/pages/Event/EventBoats.js
k3ntako: Why are we using the `MeContextProvider` if we could just do `Person.currentUser`? It comes with team and everything else?

k3ntako:
> Ugh I wish merged reviews weren't so long. Next time, just make this a branch off of your other feature branch instead of branching from master.

Sorry about that!

client/pages/Event/EventBoats.js
mienaikoe: Two Reasons:
1) Separation of concerns: Models shouldn't have to worry about the state of the application
2) Providers trigger updates: If "me" ever changes, the entire app will respond to the change, whereas you have to hardcode every response to a change to a class property on Person.

client/pages/Boat/index.js
mienaikoe: ? Confused how you arrived at that conclusion. 

Just make the renderMenu function a separate react functional component, and wrap it in MeConsumer instead of wrapping it in MeConsumer here. You shouldn't need to change any code. It's just moving where you put stuff.
Would make this easier to read.

PR #39, opened 2019-02-24
Draft.js

k3ntako:
Replaced React-Toolbox Inputs (Textarea) for Event Notes and Location Notes with Draft.js.
Rearranged Add/Edit Event page to give Location Notes the full width.

Draft.js:
- I followed an example, but removed following options: H1, H5, H6, Code Block, Block Quote, Code (inline). 
  - H5 and H6 are so small.
  - Boating events probably don't need Code Block, Block Quotes or `Code`.
  - H1 was too big, so I moved H2 to H1, H3 to H2 and H4 to H3.
- Comfortaa 700 (aka bold) isn't very bold. Not sure what we can do about this. 
  - I chose to put the notes on a Card, because the bold is a little easier to see on a white background than the blue background.


client/components/Draft-js/Editor.js
```diff
@@ -0,0 +1,109 @@
...
+
+export default class TextEditor extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      editorState: props.editorState || EditorState.createEmpty(),
+      focused: false
+    };
+    this.onChange = (editorState) => {
+      return props.updateLocationNote(editorState)
```
mienaikoe: The TextEditor shouldn't need to know what type of note it's handling, so I would rename this prop. Maybe `onChange` ?

client/components/Draft-js/Editor.css
```diff
@@ -0,0 +1,120 @@
...
+.bar::before {
+  left: 50%;
+}
+
+.bar::after {
+  right: 50%;
+}
+
+.bar.focused::before, .bar.focused::after{
+  width: 50%;
```
mienaikoe: What's the purpose of this?

client/components/Draft-js/Editor.css
```diff
@@ -0,0 +1,120 @@
...
+.styleControlsContainer{
+  position: absolute;
+  top: -2rem;
+  height: 2rem;
+  width: 100%;
+  opacity: 0;
+  pointer-events: none;
+  transition-duration: 0.35s;
+  transition-property: opacity, top;
+  transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
```
mienaikoe: oooh getting fancy with the css

client/pages/Event/EventDetails.js
```diff
@@ -1,62 +1,57 @@
...
-    return <>
-      <FontIcon className={styles.toggleNote} value={icon} onClick={this.toggleLocNote}/>
-      <div className={styles.note}>
-        {inside}
-      </div>
-    </>
+  renderLocationNote(locNote, locationName){
+    return <div className={styles.paragraph}>
+      <Card className={styles.note} style={{'fontFamily': 'Comfortaa'}}>
+        <h2 className={styles.cardTitle}>More About {locationName}</h2>
```
mienaikoe: I've had a standard of putting the section headers outside the card. If you think we should have headers inside the card, then we should talk about the benefits/downsides and each and decide on one standard.

client/components/Draft-js/Reader.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Editor } from 'draft-js';
+import DraftHelpers from './DraftHelpers'
+
+import styles from './Reader.css'
+
+export default (props) => {
+  const editorState = DraftHelpers.parseFromRaw(props.editorState)
```
mienaikoe: would be great to have a `shouldComponentUpdate` here as I imagine `parseFromRaw` is a very costly function. Just check if the props.editorState is the same. Also, it seems like editorState is a strange term for the text.

Or better yet, do this once in the calling page constructor, and don't put this call in any render function. You're putting it in a reader because you don't expect the text to change.

server/models/Event.js
```diff
@@ -18,4 +19,7 @@ module.exports = mongoose.model('Event', {
   boats: [
     { type: ObjectId, ref: 'Boat' },
   ]
-});
+}, { minimize: false });
```
mienaikoe: what does `minimize: false` do and why is it necessary? I suggest adding a comment here.

client/pages/Event/EventDetails.js
```diff
@@ -1,62 +1,57 @@
...
-    this.toggleLocNote = this.toggleLocNote.bind(this);
-    this.renderLocNote = this.renderLocNote.bind(this);
+    this.state = {};
   }
 
-  toggleLocNote(){
-    this.setState({locNoteVisible: !this.state.locNoteVisible});
+  renderEventNote(note){
+    return <div className={styles.paragraph}>
+        <Card className={styles.note} style={{'fontFamily': 'Comfortaa'}}>
```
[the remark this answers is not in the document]
k3ntako: You think it would be better to use `!important` in CSS or so it this way?

client/components/Draft-js/Editor.css
k3ntako: it's the pink bar under the input that appears when you focus.

server/models/Event.js
k3ntako: I found out that "Mongoose will, by default, "minimize" schemas by removing empty objects." When there's no text in the input, one of the objects is empty and that was causing some issues.
https://mongoosejs.com/docs/guide.html#minimize

client/pages/Event/EventDetails.js
k3ntako: Okay, should I leave this for now, and we can discuss it on Thursday?

client/components/Draft-js/Reader.js
k3ntako: Do you think I should change the name? It's just what Draft.js calls the object.

client/pages/Event/EventDetails.js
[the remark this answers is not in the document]
mienaikoe: Yea. All styles should be in CSS even if it requires !important.

client/components/Draft-js/Reader.js
mienaikoe: ah if it's a draft.js term, then no worries about the name. The rendering should still be done outside of this tho.

k3ntako: yep, I'll put it in the constructor.

client/pages/Event/EventDetails.js
mienaikoe: up to you how you want to handle it. Either way, would love to talk about it tomorrow.

k3ntako: I'll change it so it's consistent with the rest of the website. 

I want to change the `h2` to `h3` to fit in with the rest of the website. But, my worry is that it looks a little weird when the user selects a `h1` (actually `h2`). The title (`h3`) would be smaller than the text below

screen shot 2019-02-27 at 10 39 33 am
.

mienaikoe: You should provide styles specifically for headers that appear in these cards.

I would additionally only allow one kind of header. Letting people get fancy is how myspace pages got nuts.

k3ntako: I changed it so users can only have `h3` and it should be Oswald because that's the default heading font.

mienaikoe: Eh. It's a heading inside of content, so it really ought to be Comfortaa, but we can do that later.

PR #40, opened 2019-02-25
create event upsell for coach

mienaikoe:


client/pages/Team/EventsTab.js
```diff
@@ -18,6 +22,21 @@ export default function(props){
...
+    return <Link to="/events/new">
+      <Card className={styles.createEventCard}>
+        <div className={styles.createEventOverlay}>
+          <CardTitle theme={{
+              cardTitle: styles.createEventRoot,
+              title: styles.createEventTitle,
+              subtitle: styles.createEventText
+            }}
+            title="Create an Event"
+            subtitle="Start your crew's season with a practice or race"
```
k3ntako: Not great with grammar, but should there be another `a` before `race`?

k3ntako:
I LOVE THIS! Best feature so far!

client/pages/Team/EventsTab.js
mienaikoe: I think it works both ways, but I like the a

PR #41, opened 2019-02-25
Get Rid of Themr

mienaikoe:
This commit removes react-toolbox-themr in order to fix the time picker. It also removes a bunch of bells and whistles that themr had that we really don't need.


public/index.css
```diff
@@ -90,3 +90,19 @@ COMPONENTS
   margin-top: -30px;
   margin-left: -30px;
 }
+
+
+
+/* Doing what React-Toolbox-Themr failed to do */
+[data-react-toolbox=dialog]{
+  transform: translateY(0.01px) !important; /* because timepicker center calculation fails when dialog starts out 40px too high */
+  font-size: 16px; /* because the entire toolbox is assuming that font size is 16 */
```
k3ntako: Isn't this only applying to the Dialog? If all React Toolbox components assume `font-size: 16px` shouldn't we have something that applies to all?

public/index.html
```diff
@@ -5,8 +5,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>Mighty Speed Krush</title>
   <link rel="stylesheet" href="/index.css" />
-  <link rel="stylesheet" href="/react-toolbox/theme.css" />
-  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Comfortaa:400,700|Oswald:300,400,500,600,700|Material+Icons">
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Comfortaa:400,700|Oswald:300,400,500,600,700|Material+Icons|Roboto:400">
```
k3ntako: Don't think you're using Roboto, but I assume we are going to?

public/index.css
mienaikoe: AH yes! I forgot that only targeted dialogs

public/index.html
mienaikoe: I briefly tried this as a solution to the time picker problem, and accidentally left it. I'll revert

mienaikoe:
aight we'll check it out during the mtg

PR #42, opened 2019-02-28
Form Validation

k3ntako:


client/pages/Auth/Join/JoinPage.js
```diff
@@ -51,13 +57,19 @@ class JoinPage extends React.Component {
...
       this.setState({
-        error: error,
+        error: error.message,
         fetching: false
       });
     })
   }
 
   submit(){
+    const EMAILREGEX = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
mienaikoe: You can move this out of the class definition to prevent creating it for every submit.

server/controllers/BoatController.js
```diff
@@ -8,23 +8,35 @@ const Boat = require('../models/Boat')
...
-          boats: boatMongoose
+    Event.findById(req.params.event_id).
+      populate("boats").
+      then(event => {
+        existing = event.boats.find(boat => {
+          return boat.name === req.body.name
+        })
+
+        if(existing){
+          throw new ValidationError();
```
mienaikoe: Include a message in your errors

server/controllers/BoatController.js
```diff
@@ -8,23 +8,35 @@ const Boat = require('../models/Boat')
...
-    }).catch( next )
+
+        return event;
+      }).then(event => {
+        let boat = new Boat({
+          _id: new mongoose.Types.ObjectId(),
+          name: req.body.name,
+        });
+
+        let boatRet;
```
mienaikoe: you can prevent the need for this variable by nesting the promise below:
```
return boat.save().then( boatMongoose => {
     event.boats.push(boatMongoose)
     return event.save().then( () => {
          res.json(boatMongoose);
          next();
    }).catch( next )
```

There are upsides and downsides to doing this, but consider it and make a decision which one you prefer and know why.

server/controllers/PersonController.js
```diff
@@ -38,65 +38,71 @@ const FIND_POPULATOR = {
...
+          req.session.person_id = person._id
+          req.session.team_id = person.team;
+
+          return person.populate(DEEP_POPULATOR).execPopulate();
+        }).then( person => {
+          person.set('auth', null)
+          res.json(person);
+          next();
+        }).
+        catch( err => {
```
mienaikoe: can be changed to `catch( next )`

server/models/Boat.js
```diff
@@ -28,7 +28,7 @@ const boatSeatSchema = new mongoose.Schema({
 
 module.exports = mongoose.model('Boat', {
   _id: ObjectId,
-  name: String,
+  name: { type: String, required: true, trim: true, minlength: 1 },
```
mienaikoe: Wow! I didn't know trim: true was a thing!

server/controllers/BoatController.js
k3ntako: I think it's easier to read the way it is, so I will leave it.

PR #44, opened 2019-03-01
Changed tab change behavior

k3ntako:


client/pages/Boat/index.js
```diff
@@ -43,7 +43,7 @@ class BoatPage extends Component {
       let index = event.boats.findIndex( boatIt => boatIt._id === this.props.match.params.id );
 
       if(index === -1){
-        return this.props.history.push(
+        return this.props.history.replace(
```
k3ntako: Do you think this should be a `push` or a` replace`?

I chose `replace` because I didn't want to create a loop where the 1) user is redirected, 2) hits back, and 3) then redirected again.

PR #46, opened 2019-03-07
Offline Mode

k3ntako:


public/workbox/sw.js
```diff
@@ -0,0 +1,96 @@
...
+      const blob = new Blob([JSON.stringify({isOnline: false}, null, 2)], {type : 'application/json'});
+
+      return new Response(
+        blob,
+        // {status: 599} //issue with this is that it throws an error in Fetcher
+      );
+    }
+  });
+
+}
```
k3ntako: move sw.js to client

public/workbox/sw.js
```diff
@@ -0,0 +1,96 @@
...
+          statuses: [0, 200],
+        })
+      ]
+    })
+  );
+
+  workbox.routing.registerRoute(
+    /\/api\/\b(workbox)\b/,
+    new workbox.strategies.NetworkOnly()
+  );
```
k3ntako: try navigator.online?

public/workbox/sw.js
```diff
@@ -0,0 +1,96 @@
...
+
+  workbox.routing.registerRoute(
+    /^https:\/\/maps\.(googleapis|gstatic)\.com.*$/,
+    new workbox.strategies.NetworkOnly()
+  );
+
+  //Default Handler (if no other matches)
+  workbox.routing.setDefaultHandler(({url, event, params}) => {
+    console.log("Not cached: ", event.request.method, event.request.url)
+  });
```
k3ntako: maybe hide it in production?

public/workbox/sw.js
```diff
@@ -0,0 +1,96 @@
...
+
+  //Error Handling
+  workbox.routing.setCatchHandler(({event}) => {
+    if(event.request.url.includes("/api/workbox/isonline")){
+      const blob = new Blob([JSON.stringify({isOnline: false}, null, 2)], {type : 'application/json'});
+
+      return new Response(
+        blob,
+        // {status: 599} //issue with this is that it throws an error in Fetcher
```
k3ntako: uncomment, and catch in model

client/models/Workbox.js
```diff
@@ -0,0 +1,15 @@
...
+
+  }
+
+  static isOnline(){
+    return Fetcher.get('/api/workbox/isonline').then( res => res.isOnline )
+  }
+  
+}
+
+export default Workbox
```
k3ntako: put it in utilities

client/components/Nav/Nav.js
```diff
@@ -22,6 +23,23 @@ class Nav extends Component {
...
+
+  startNetworkCheck(){
+    setInterval(() => {
+      let { isOnline, networkChecked } = this.props;
+      if(!isOnline || !networkChecked || this.props.sinceLastCheck > 30){
+        this.props.checkNetwork();
+      }else{
+        this.props.addCheckCount();
+      }
+    }, 10000);
```
k3ntako: addeventlistener with ononline

client/service-worker.js
```diff
@@ -0,0 +1,93 @@
...
+  );
+
+  workbox.routing.registerRoute(
+    /^https:\/\/maps\.(googleapis|gstatic)\.com.*$/,
+    new workbox.strategies.NetworkOnly()
+  );
+
+  //Default Handler (if no other matches)
+  workbox.routing.setDefaultHandler(({url, event, params}) => {
+    console.log("Not cached: ", event.request.method, event.request.url)
```
k3ntako: maybe hide it in production?

k3ntako:
Added Workbox to cache most API calls, fonts, and static files
- Allows users to load pages that they have already loaded.
Redux keeps track of whether user is connected to the internet and our servers.
- After being disconnected from either, if the user is reconnected the page will reload.
- Reason for reload: if API call isn't cached the information will not load on page even if the component receives new props, because fetch to API is generally made in `componentDidMount()`.

Plan: convert all providers to Redux.

client/service-worker.js
mienaikoe: you could use console.debug instead. This really isn't that big a deal

client/pages/Event/index.js
```diff
@@ -73,16 +75,18 @@ class EventPage extends Component {
     const event = this.state.event;
     const me = meState.me;
     if( me.isCoach() ){
+      const editMenu = <MenuItem value='edit' icon='edit' caption='Edit Event' disabled={!this.props.isOnline}/>;
+      const menuItem = this.props.isOnline ? <Link to={`/events/${event._id}/edit`}>{editMenu}</Link> : editMenu;
```
mienaikoe: This was a bit hard to read because the `isOnline` check is in two places. You could just use a single if/else on `this.props.isOnline` and define the menuItem differently based on that.

client/pages/EventForm/EventMap.css
```diff
@@ -25,3 +25,28 @@
...
+.disabledMap > h3{
+  text-align: center;
+  margin-top: 0;
+}
+
+.offlineIcon{
+  font-size: 2rem;
+  margin-right: 1rem;
+  color: var(--palette-deep-orange-500);
+  background-image: radial-gradient(circle, white 44%, transparent 44%);
```
mienaikoe: The background actually makes sense here because the empty map would make the icon hard to see. Couldn't you just use `border-radius` and `background-color` to make the background circle, tho? Not saying this way is any better or worse, but it seems hacky because there's not actually a gradient as the function would suggest. It just happens to be a circle because radial-gradients are circular.

client/pages/Team/EventCard.js
```diff
@@ -56,7 +57,12 @@ export default class EventCard extends Component {
     } else if( this.props.event.is_attending(meState.me) ){
       return <Button raised disabled> Attending </Button>
     } else {
-      return <Button raised primary onClick={this.attend.bind(this, true)} primary>
+      return <Button
+        raised
+        primary
+        onClick={this.attend.bind(this, true)}
+        primary
```
mienaikoe: primary is here twice

client/pages/Team/EventsTab.js
```diff
@@ -23,20 +24,23 @@ function EventsTab(props){
...
-              title: styles.createEventTitle,
-              subtitle: styles.createEventText
-            }}
-            title="Create an Event"
-            subtitle="Jumpstart your crew's season with a practice or a race"
-          />
-        </div>
-      </Card>
-    </Link>
+    const imageTitle = props.isOnline ? "Create an Event" : "Exciting Things are Ahead"
```
mienaikoe: tbh, I wouldn't even bother rendering the cards if the user is offline. There's no way to show that they're disabled, and it would be misleading. You can just render an `<Error/>` that says the user is offline.

client/redux/modules/workbox.js
```diff
@@ -0,0 +1,56 @@
...
+    case UPDATE_NETWORK_STATUS:
+      return Object.assign({}, state, {
+        isOnline: navigator.onLine,
+      })
+    default:
+      return state
+  }
+}
+
+const START_SERVER_CHECK = 'START_SERVER_CHECK';
```
mienaikoe: I suggest defining these at the top.

client/utilities/Workbox.js
```diff
@@ -0,0 +1,53 @@
...
+
+  setTimeout(() => {
+    expBackoff(delay)
+  }, delay)
+}
+
+const fetchServer = () => {
+  return fetch('/api/workbox/server-status', {
+    method: "GET",
+    credentials: "include",
```
mienaikoe: why do you need a cookie for this?

client/utilities/Workbox.js
```diff
@@ -0,0 +1,53 @@
...
+    return;
+  }
+
+  fetchServer();
+
+  if(delay >= MAX_MILLISECONDS){
+    return;
+  }
+
+  delay = delay * 2 < MAX_MILLISECONDS ? delay * 2 : MAX_MILLISECONDS;
```
mienaikoe: You can replace this with `Math.min(delay * 2, MAX_MILLISECONDS)`

It should be more performant and easier to read.

client/utilities/Workbox.js
```diff
@@ -0,0 +1,53 @@
...
+const expBackoff = (delay) => {
+  let { isOnline, isServerConnected } = store.getState().workbox;
+  if(!isOnline || isServerConnected){
+    return;
+  }
+
+  fetchServer();
+
+  if(delay >= MAX_MILLISECONDS){
+    return;
```
mienaikoe: Why not just keep the delay at the max, and keep trying. Doing something once every 20 minutes is not a big deal. Some train rides go well over 20 minutes for those in brooklyn.

package.json
```diff
@@ -44,6 +46,7 @@
     "css-loader": "^2.1.0",
     "mocha": "^5.2.0",
     "style-loader": "^0.23.1",
-    "webpack-cli": "^3.1.2"
+    "webpack-cli": "^3.1.2",
```
mienaikoe: Does webpack require this now? I thought it came with regular webpack.

webpack.config.js
```diff
@@ -36,6 +37,14 @@ var config = {
...
     ]
   },
+	plugins: [
+    new InjectManifest({
+      swSrc: APP_DIR + '/service-worker.js',
+      swDest: BUILD_DIR + '/service-worker-build.js',
+			exclude: [/[^]*/],
+			// importsDirectory: BUILD_DIR + "/workbox/build/precache-manifest/"
+    })
+	],
```
mienaikoe: Some weird spacing in this section

client/utilities/Workbox.js
mienaikoe: I just realized this is for server outage, not offline. Still would like to have the app keep trying. Maybe even more often than 20 minutes.

client/components/Nav/index.css
```diff
@@ -35,3 +35,16 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+
+}
+
+.offline{
+  color: var(--palette-deep-orange-500);
+  padding-right: 3px;
+  background-image: radial-gradient(circle, white 44%, transparent 44%);
```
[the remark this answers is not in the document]
k3ntako: The center of the icon is see through. This just makes it so the center "bolt" is always white.

client/index.js
```diff
@@ -4,26 +4,59 @@ import { BrowserRouter as Router } from 'react-router-dom';
...
+          </MeProvider>
+        </Provider>
       </DragDropContextProvider>
     </Router>
   }
 }
 
 render(<App/>, document.getElementById('app'));
+
+export { store };
```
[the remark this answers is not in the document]
k3ntako: So, the only reason I export `store` is to use Redux in non-React components (i.e. Model and utilities). 

And yeah `mapDispatchToProps` does work for all the components, but Models/Utilities don't have props? I guess I could pass the function through, but that doesn't seem better? 

I read something about using Middlewares, but I thought they still required the use of `dispatch()`, which is only available using `store.dispatch` or `connect` (only available for components).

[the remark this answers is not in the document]
k3ntako: I thought about making `Fetcher` into a class but I realized that the REST actions would be `static` and wouldn't have access to `props`.

client/pages/Event/index.js
k3ntako: I'm a bit confused about what you wanted. Does either look like what you were thinking?

```
let menuItem = <MenuItem
  value='edit'
  icon='edit'
  caption='Edit Event'
  disabled={!this.props.isOnline} />;

if(this.props.isOnline){
  menuItem = <Link to={`/events/${event._id}/edit`}>{menuItem}</Link>
}
```

```
let menuItem;
  if(this.props.isOnline){
    menuItem = <Link to={`/events/${event._id}/edit`}>
      <MenuItem
        value='edit'
        icon='edit'
        caption='Edit Event'
        disabled={false} />
    </Link>
  }else{
    menuItem = <MenuItem
      value='edit'
      icon='edit'
      caption='Edit Event'
      disabled={true} />;
  }
```

client/pages/EventForm/EventMap.css
k3ntako: I tried `border-radius` but the white background would be bigger than the icon. Currently, I just want the center of the icon to have a white background, so the see through part is consistently white, instead of the color behind it.

client/components/Nav/index.css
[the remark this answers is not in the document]
mienaikoe: Ohhhhh ok. I get it now.

mienaikoe:
Still has merge conflicts

mienaikoe:
conflicts fixed. Merging.

client/index.js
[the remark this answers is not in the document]
mienaikoe: So I merged, this but I want to revisit this next time we meet.

Models and Utilities should never introspect into the state of the application. React, and pretty much any other framework you'll work with maintains this idea of a single direction of dependencies. That means the app should depend on Models, and the Models should depend on the API. This also goes for the app depending on utilities that sometimes depend on models.

This is important for two reasons:
- When you create clean divisions in your code, it makes it easier to separate sections of it into different libraries for use in other codebases.
- When you have one dependency direction, anyone who hasn't read your code before can follow your code far more easily. The opposite of this is called spaghetti code, and it makes debugging or refactoring your code far more tedious and frustrating.

PR #47, opened 2019-03-12
Email Notifications

mienaikoe:
- Refactored a lot of the email and setup code
- Introducing email-notifications library
- Creating Templates for emails (html-escaping, clarity of file structure, etc)

Note: This is a commit that defines the task. More work will be needed to use heroku scheduler to actually run the task on a schedule.


server/tasks/event_reminder.js
```diff
@@ -0,0 +1,100 @@
...
+
+
+
+const nowMillis = new Date().getTime();
+const ONE_DAY_IN_MS = 1000 * 60 * 60 * 24;
+
+Event.find({
+  datetime: {
+    // at least one day in advance. Any less is not enough time for the coach to respond to
+    $gt: new Date( nowMillis + ONE_DAY_IN_MS ),
```
k3ntako: let's say we do this at 8am, wouldn't it still be worth including that day's event?

server/tasks/event_reminder.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+}).populate({
+  path: 'team',
+  model: 'Team',
+  populate: {
+    path: 'members',
+    model: 'Person',
+  }
+}).then( events => {
+  console.log(`${events.length} Events found. Sending event reminders.`)
```
k3ntako: repetitive? see above.

server/utilities/reset_password.js
```diff
@@ -0,0 +1,36 @@
+require('dotenv').config();
+const nodemailer = require("nodemailer");
+const crypto = require('crypto');
+const Email = require('email-templates');
+const path = require('path');
+
+let mailConfig = config.mailer;
```
k3ntako: Why don't you need: `const config = require('../../config.js')[process.env.NODE_ENV];`

package.json
```diff
@@ -2,25 +2,28 @@
...
   "scripts": {
-    "test": "mocha \"test/**/*.js\"",
-    "dev": "concurrently --raw \"webpack --mode=development --watch\" \"NODE_ENV=development node server.js\"",
-    "migrate": "node migrations/run.js",
-    "heroku-postbuild": "webpack -p"
+    "test": "mocha \"server/tests/*.js\"",
+    "dev": "concurrently --raw \"webpack --mode=development --watch\" \"NODE_ENV=development node server/index.js\"",
+    "migrate": "node server/migrations/run.js",
+    "heroku-postbuild": "webpack -p",
+    "event-reminders": "NODE_ENV=production node server/tasks/event_reminder.js"
```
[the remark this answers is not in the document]
mienaikoe: so heroku scheduler will run this on a schedule. I'm thinking we'll do it every day at like midnight.

server/index.js
```diff
@@ -9,8 +9,9 @@ const path = require('path');
 // EXPRESS SETUP
 const app = express();
 const environment = app.get('env')
-const config = require('./config.js')[environment];
-const apiRoutes = require('./server/routes');
+const config = require('./config')[environment];
+global.config = config;
```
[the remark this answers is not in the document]
mienaikoe: `global` is a global variable in node.js that is sort of akin to `window` in the browser.

adding config to `global` means that any file can access `global.config` without having to import anything. I forget exactly why I needed this...

server/tasks/event_reminder.js
```diff
@@ -0,0 +1,100 @@
...
+
+
+/*
+  Send reminder Emails for an event
+
+  Fetches every person on the team, and finds if they haven't responded yet.
+  Every non-responder gets BCC'd to a reminder email
+*/
+function sendReminderEmail( event ){
+  console.log(`Sending Reminder Email for ${event.name}`);
```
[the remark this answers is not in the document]
mienaikoe: console.log is fine for a scheduled task.

Generally, most larger projects will use a logger like log4js on the backend. It writes to a file in a non-blocking way and has lots of formatting things.

For frontend, errors should really just be thrown unless there's a legit way to recover from them. Console logging isn't really appropriate on frontend because it slows down your app and the output will likely be worthless anyway since your code is minified. Most people will just log errors and then send them to some log aggregator like Splunk.

For this project... It's like small enough that it probably doesn't matter. If I bring it up in code review, it's because the logging probably isn't useful anyway. In this case, a scheduled task has no other way of leaving a trail in case it fails, so I opted to log its behaviors.

server/tasks/event_reminder.js
mienaikoe: No, because it will already have been reminded about yesterday.

server/tasks/event_reminder.js
mienaikoe: No: if there are zero events, I want to know. I also want to know which events are being sent in case the selection logic gets it wrong, or we need to diagnose why a certain event's emails didn't go through.

server/utilities/reset_password.js
mienaikoe: because of the above change in server/index.js

But seeing as that's kind of not actually needed, I can change this back to using env.

server/index.js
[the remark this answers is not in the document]
mienaikoe: yea I'll remove it

server/tasks/event_reminder.js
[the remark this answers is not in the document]
k3ntako: Thanks for all these long responses!

k3ntako:
Thanks for all the responses! Could you resolve the merge conflicts?

mienaikoe:
conflicts resolved

mienaikoe:
Also, I would suggest that we don't merge this until after your offline stuff is in.

k3ntako:
Noticed that `email` variable was used for two different things in the `reset_password` utility. I pushed an updated to fix it.

mienaikoe:
> Noticed that `email` variable was used for two different things in the `reset_password` utility. I pushed an updated to fix it.

Thanks for catching that!

PR #49, opened 2019-03-19
Adding Team to Breadcrumbs

mienaikoe:
- On event, event form, series form, and boat pages, adds a team breadcrumb
- New CSS for mobile team page to have a better fit for text


client/pages/Team/EventCard.css
```diff
@@ -44,3 +47,20 @@
     right: 0;
     bottom: 0;
 }
+
+
+@media screen and (max-width: 600px){
+  .datetime{
+    font-size: 24px;
```
k3ntako: The datetime becomes two lines on mobile. Was that intended? Also, it's not consistent between event cards (see image)


Example:
```
11:09
PM
```

Screen Shot 2019-03-19 at 1 53 30 PM

client/pages/Team/EventCard.css
```diff
@@ -22,16 +22,19 @@
 .cardTitle{
   font-family: 'Oswald', sans-serif;
   font-size: 28px !important;
+  white-space: nowrap;
+  text-overflow: ellipsis;
```
k3ntako: `ellipsis` doesn't seem to be working.

Screen Shot 2019-03-19 at 1 53 30 PM

client/pages/Team/EventCard.css
mienaikoe: Yea it was intended for mobile, because it saves space for the title.

I'll address the long title issue.

PR #51, opened 2019-03-19
Fancy Buttons

mienaikoe:
- Changed the design of buttons to be more inline with modern material design
- Changed the attendance buttons on event card to be yes/no
- Added tooltips to the attendance buttons


client/components/Button/theme.css
```diff
@@ -0,0 +1,18 @@
...
+  font-family: 'Comfortaa', sans-serif;
+  font-weight: 700;
+  font-size: 1rem;
+  letter-spacing: 0.1rem;
+  padding: 0 1rem;
+}
+
+.button:disabled {
+  background-color: var(--palette-grey-200);
+  color: var(--palette-grey-500)
```
k3ntako: missed a semicolon

client/pages/Team/CreateEventCard.css
```diff
@@ -1,15 +1,22 @@
...
   filter: grayscale(0.5);
   cursor: pointer;
 }
 
 .createEventOverlay{
-  padding: 4rem;
-  background-color: rgba(0,0,0,0.5);
+  padding: 1rem 3rem;
+  background: linear-gradient(to top right, var(--palette-blue-grey-900), transparent );
+  /* background-position: 0 100%; */
```
k3ntako: comment

client/pages/Team/CreateEventCard.css
```diff
@@ -19,26 +26,56 @@
 .createEventRoot .title,
 .createEventRoot .subtitle{
   color: white;
-  text-shadow: 0px 0px 12px rgba(0,0,0,0.5);
-  text-align: center;
+  /* text-shadow: 0px 0px 12px rgba(0,0,0,0.5); */
+  /* text-align: center; */
```
k3ntako: comment

client/pages/Team/EventCard.css
```diff
@@ -35,12 +32,41 @@
...
     right: 0;
     bottom: 0;
 }
+.action{
+  background-color: white;
+  color: var(--palette-grey-800);
+  margin-left: 1rem;
+}
+.action.active{
+  color: white !important;
```
k3ntako: I think these buttons should have a border or a shadow. I think the inactive button is hard to tell that it's a button. The "not attending" button is a line in white space.

client/pages/Team/CreateEventDialog.css
```diff
@@ -6,7 +6,7 @@
 
 .miniCard{
```
k3ntako: this is a little unrelated, but these `miniCard`s should stack on top of each other on mobile. 

Also, in general, I feel like dialogs could be a little wider by default.

Screen Shot 2019-03-19 at 6 30 05 PM

client/pages/Team/CreateEventCard.css
```diff
@@ -19,26 +26,56 @@
...
+  transition: opacity 0.3s ease;
+}
+.createEventRoot:hover .title::after{
+  opacity: 1;
+}
 
 .createEventRoot .title{
-  font-weight: bold !important;
-  font-size: 2.7rem !important;
+  font-family: Oswald !important;
```
k3ntako: should `Oswald` be in quotes?

client/pages/Team/CreateEventCard.css
```diff
@@ -19,26 +26,56 @@
...
+  font-weight: 300 !important;
+  font-size: 3.5rem !important;
 }
 
 .createEventRoot .subtitle{
+  font-family: Oswald !important;
```
k3ntako: same as above

client/pages/Team/EventCard.js
```diff
@@ -18,52 +28,75 @@ class EventCard extends Component {
...
+        <div key="attendees">
+          <span> {event.attendees.length} </span>
+          <label> Attending </label>
+        </div>
+        <div key="responses">
+          <span> {event.attendees.length + event.non_attendees.length} </span>
+          <label> Responded </label>
+        </div>
+      </div>
+    } else if( event.datetime < new Date() ){
```
k3ntako: I thought it was a little strange that the event was on the home page as upcoming but you couldn't respond to it. What if we made this act the same way as the events themselves (ie you can respond until midnight of that day)?

client/components/Button/theme.css
```diff
@@ -0,0 +1,18 @@
+
+.button{
+  font-family: 'Comfortaa', sans-serif;
```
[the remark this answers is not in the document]
mienaikoe: Yea. Comfortaa all caps was much more legible and good looking.

client/pages/Team/EventCard.css
mienaikoe: I'll add a shadow and change the icon

client/pages/Team/CreateEventDialog.css
mienaikoe: Oh jeez.

client/pages/Team/CreateEventCard.css
mienaikoe: yes

client/pages/Team/EventCard.js
mienaikoe: Why would you want to respond after the event?

k3ntako: So, I feel like sometimes I over emphasize worst case scenarios. 

I'm thinking that someone who wasn't planning to come text coach to come and doesn't respond on the app. After the event starts (before it ends), coach wants to change the lineup but can't add this member because they responded that they aren't coming. There is no way for coach to change and share the lineup and see the weight balance.

mienaikoe: Hmm I was also considering this case for filling out our team with people from Empire or h2o. Or even unsetting a seat, because we don't currently have an option for that.

I think we need to provide an option for a write-in, where you just add a weight to a seat, without a name. What do you think?

mienaikoe: Also, I was just refactoring this logic to be easier to read - I didn't change the signup cutoff logic. I don't think this PR should cover changes to event signup logic, as it's mostly about the buttons.

client/pages/Team/EventCard.js
```diff
@@ -18,52 +28,79 @@ class EventCard extends Component {
...
-        raised
-        primary
-        onClick={this.attend.bind(this, true)}
-        disabled={!this.props.isOnline}>
-        Attend
-      </Button>
+      let response = event.findResponseForPerson(me);
+      return [
+        <TooltipButton key='negative'
+          icon='format_color_reset'
```
k3ntako: see above

client/pages/Event/index.js
```diff
@@ -109,16 +109,23 @@ class EventPage extends Component {
     } else if( event.datetime > new Date() ) {
       let response = event.findResponseForPerson(me);
       return <div className={styles.eventActions}>
-        <RippleButton
-          icon='clear'
-          floating primary={response === false}
+        <TooltipButton
+          icon='remove'
```
k3ntako: You used `remove` here but you used `format_color_reset` on the `EventCard`, see below. Shouldn't we keep the icons consistent?

mienaikoe: yea. I'll switch this to format_color_reset

PR #52, opened 2019-03-22
Send Team Emails

k3ntako:
Allows any user to send an email to the team.
- It's assumed that you are emailing every member unless they are exclude.
- Can add non-team members.

MSK Admin is part of our team. What should we do with that account?


client/pages/Email/ExcludeForm.js
```diff
@@ -0,0 +1,167 @@
...
+
+    return <Card className={styles.card}>
+      <div className={rStyles.formTitle}>Exclude a Member from Email</div>
+      <p>
+        Email will be sent to all coaches and members, however, you can choose to exclude any recipients.
+        <br /><br />
+        You can include a member that was previously excluded by selecting them in the same text box below.
+      </p>
+
+      <Input
```
mienaikoe: Looks like this is an autocomplete. Any thoughts on using a Table instead? I see some pros and cons to each, but thought I'd open up a discussion about it.

public/index.css
```diff
@@ -5,6 +5,7 @@
 
 html{
   font-size: 12px;
+  height: 100%;
```
mienaikoe: No! This constrains the document to be the viewport height. Scrolling gets all out of wack. Why is this here?

server/models/Email.js
```diff
@@ -0,0 +1,36 @@
...
+var EmailSchema = new Schema({
+  _id: ObjectId,
+  sender: { type: ObjectId, ref: 'Person', required: true },
+  subject: { type: String, trim: true },
+  team: { type: ObjectId, ref: 'Team', required: true, index: true },
+  exclude: [
+    { type: ObjectId, ref: 'Person' }
+  ],
+  nonMember: [
+    { type: Object }
```
mienaikoe: Let's make a schema for nonMembers instead of a generic object

server/models/Email.js
```diff
@@ -0,0 +1,36 @@
...
+  ],
+  emailBody: {
+    type: Object,
+    required: true,
+    default: draftRaw
+  },
+  sendTime: { type: Date, default: null},
+  rejected: [
+    { type: String }
```
mienaikoe: Would boolean suffice here?

server/models/Email.js
k3ntako: so, this is a list of emails that were "rejected" (I think that means bounced back)

server/models/Email.js
k3ntako: this is a list of emails that were "rejected" (I think that means bounced back)

mienaikoe: Ahh ok. so then the purpose is to tell coach which email didn't get received. Aight that's fine.

server/tests/TestHelper.js
```diff
@@ -68,6 +71,31 @@ module.exports = {
...
+        body: {
+          first_name: 'Team',
+          last_name: 'Mate',
+          email: email || `TeamMate${randomnum()}@wakeBird.com`,
+          password: "testtesttest",
+        },
+      }
+      const res = {
+        json: function(response){
+          test_teammate = response;
```
mienaikoe: I want to avoid any more saving of variables on this file. The variables should be saved in their respective tests, because the order and sequence of the tests can't be guaranteed, and the tests shouldn't rely on each other to have already called certain functions.

This isn't documented anywhere, so I don't blame you. We can talk about it a bit more on Monday. but for now, please remove test_teammate from this file, and just use the response from this function.

server/utilities/team_email.js
```diff
@@ -0,0 +1,56 @@
...
+
+  return accumulator;
+};
+
+const recipientList = (teamEmail) => {
+  const members = teamEmail.team.coaches.concat(teamEmail.team.members).concat(teamEmail.nonMember);
+  const exclude = teamEmail.exclude.map(person => person.email);
+
+  //remove duplicate emails and exclude members
+  const recipients = members.reduce((accumulator, currentValue) => {
```
mienaikoe: This isn't the optimum way to create the recipients list. This is the sort of thing that will be in an interview. Your challenge is to find a quicker way to do this.

client/pages/Team/EmailDraftTable.js
```diff
@@ -0,0 +1,110 @@
...
+      return <Error
+        errorTitle={"No Drafts!"}
+        message={composeButton}
+        icon="inbox"
+        />
+    }
+
+    return <>
+      <div className={styles.draftHeading}>
+        <h2>Drafts</h2>
```
mienaikoe: I recommend using h3 instead of h2

client/pages/Team/EmailTableTemplate.js
```diff
@@ -0,0 +1,57 @@
...
+    tableHead = <TableHead>
+      <TableCell>Subject</TableCell>
+      <TableCell>Sent</TableCell>
+    </TableHead>
+  }else{
+    tableHead = <TableHead>
+      <TableCell className={styles.checkBox}>
+        <Checkbox
+          checked={props.checkAll}
+          onChange={() => props.handleChange('checkAll')} />
```
mienaikoe: This didn't seem to work when I tested it locally. It just stayed unchecked when I clicked it on drafts.

client/pages/Team/EmailDraftTable.js
```diff
@@ -0,0 +1,110 @@
...
+        <div className={styles.buttons}>
+          <div>
+            <Button
+              raised primary
```
mienaikoe: Doesn't need to be raised

client/pages/Team/EmailDraftTable.js
```diff
@@ -0,0 +1,110 @@
...
+        errorTitle={"You're Offline!"}
+        message={"Check back when you're back online"}
+        icon="offline_bolt"
+        />
+    }
+
+    const composeButton = <Button
+      raised primary
+      label="Compose"
+      onClick={this.compose.bind(this)} />
```
mienaikoe: you could make this an icon+text button to give it some visual hook. Same with the delete button.

client/pages/Email/index.js
```diff
@@ -0,0 +1,184 @@
...
+          index={tabIndex}
+          onChange={this.handleTabChange.bind(this)}
+          theme={{
+            tab: styles.tab,
+            tabs: styles.tabs
+          }}>
+          <Tab label='Email' className={styles.tabHeader}>
+            <div className={styles.spaceBetweenButton}>
+              <div className={styles.link}>
+                <Link to="/#email">Back to Inbox</Link>
```
mienaikoe: 1) It's not an inbox if the list is all outgoing mail
2) This is the sort of thing crumbs in the header are for. If you think that it takes up so much space that this is the only place in the whole site that the header shouldn't be, then perhaps we should shrink the header a bit. Let's talk about this Monday.

client/pages/Email/index.js
```diff
@@ -0,0 +1,184 @@
...
+              subjectChangeHandler={this.subjectChangeHandler.bind(this)}
+              isSending={isSending}
+              setIsSending={this.setVerb.bind(this, "isSending")}
+              isSaved={isSaved}
+              setIsSaved={this.setIsSaved.bind(this)}
+              isSaving={isSaving}
+              setIsSaving={this.setVerb.bind(this, "isSaving")}
+              /> }
+          </Tab>
+          <Tab label='Recipients' className={styles.tabHeader}>
```
mienaikoe: I completely missed the recipients tab the first couple times I was using this. Maybe it shouldn't be in tabs. I get that it's a relatively large set of options, but perhaps most of them can be put into modals instead of cards that fill up the page space.

client/pages/Email/ExcludeForm.js
mienaikoe: Also, I have a more general case of a person autocomplete coming up in the assignments branch that you can use instead of this if you still think autocomplete is the right way to go.

client/pages/Email/ExcludeForm.js
```diff
@@ -0,0 +1,167 @@
...
+  include(){
+    this.props.email.excludeMember(this.state.selectedPerson.id, false).then(() => {
+      this.setState({search: "", selectedPerson: null}, this.props.forceUpdate)
+    });
+  }
+
+  render(){
+    const { isOnline, isServerConnected } = this.props;
+
+    const disableButtons = !selectedPerson || !isOnline || isServerConnected;
```
mienaikoe: selectedPerson isn't defined until line 104. This makes `disableButtons` disabled forever

client/pages/Email/NonMemberForm.js
```diff
@@ -0,0 +1,142 @@
...
+          onClick={this.add.bind(this)}
+          disabled={disableButtons || !this.valid()}/>
+      </div>
+
+      <div className={rStyles.formTitle} id={rStyles.removeNonMember}>
+        Remove Non-Member from Email
+      </div>
+      <p>Remove recipients that were added above.</p>
+      <div className={rStyles.inputRow}>
+        <Input
```
mienaikoe: this should be done via a button. There's no need to re-enter information that the app already has.

public/index.css
k3ntako: Put buttons at bottom and make Draft.js the height of content.

k3ntako: make delete button an icon

server/utilities/team_email.js
k3ntako: don't need to exclude `nonMembers`

use filter for exclude and take it out of reducer

client/pages/Email/index.js
k3ntako: Add member modal

exclusion inclusion and deletion of non member can be done on list

list can be under editor

server/utilities/team_email.js
```diff
@@ -8,24 +8,11 @@ const convertFromRaw = require('draft-js').convertFromRaw;
...
-  if(!accumulator.includes(address) && !exclude.includes(address)){
-    accumulator.push(address)
-  }
-
-  return accumulator;
-};
-
 const recipientList = (teamEmail) => {
-  const members = teamEmail.team.coaches.concat(teamEmail.team.members).concat(teamEmail.nonMember);
+  const members = teamEmail.team.members.map(person => person.email);
```
k3ntako: I wasn't too sure about this, but coaches are members, right?

mienaikoe: In the migrations I specified that coaches were also members. So this should be ok.

PR #54, opened 2019-03-23
Event Page Redesign

mienaikoe:
- Put smaller information tidbits in a sidebar
- Divided Event Note from Location Details/Map
- Added Google Calendar Link


client/pages/Event/index.css
```diff
@@ -14,29 +14,36 @@
...
 }
 
-.cardTitle{
-  padding-left: 16px;
-  margin-bottom: 1rem;
+.locationNote{
+  margin-top: 2rem;
+  border-top: 1px solid var(--palette-grey-200);
+  padding-top: 2rem;
+}
```
k3ntako: If there's no note, it might be better to not have a border?

client/models/Location.js
```diff
@@ -29,6 +29,21 @@ class Location extends Model {
...
+    const firstComma = this.address.indexOf(",");
+    if( firstComma ){
+      return [
+        this.address.substring(0, firstComma).trim(),
+        this.address.substring(firstComma+1).trim()
+      ];
+    } else {
+      return [this.address]
+    }
+  }
```
k3ntako: Don't think this will be an issue, but if you put in a city name as your location it gets a little weird. I doubt anyone would that. Just wanted to point this out.

If you think this is an issue, we could save more detailed info for address, like an object with street address, city, state, zip, etc.

For example, Hartford, CT would 
```
Hartford
CT, USA
```

client/pages/Event/index.css
mienaikoe: I'll make sure the element only gets rendered if there's a note

client/models/Location.js
mienaikoe: Eh. It's not a case I really care about, and that doesn't look terrible at least.

PR #55, opened 2019-03-26
Making Sex Field Coach-Only

mienaikoe:


client/pages/Person/PersonPage.js
```diff
@@ -70,15 +72,14 @@ class PersonPage extends Component {
...
-            <Stat label="Shirt Size" value={person.shirt} />
-            <Stat label="CPR Certified" value={person.cpr_cerified} />
-          </Card>
+          <Table selectable={false}>
+            { this.renderStat("Email", person.email) }
+            { this.renderStat("Phone", person.phone) }
+            { this.renderStat("Birthday", EventFormatter.formatBirthday(person.birthday)) }
+            { this.renderStat("Neighborhood", person.location) }
+            { this.renderStat("Shirt Size", person.shirt) }
+            { this.renderStat("CPR Certified", person.cpr_cerified) }
```
k3ntako: I think this needs to break twice or at a higher breakpoint. The month gets cut off around 720px.

Screen Shot 2019-03-27 at 5 40 12 PM

client/pages/Person/PersonPage.js
```diff
@@ -43,6 +38,13 @@ class PersonPage extends Component {
     });
   }
 
+  renderStat(label, value){
+    return <TableRow className={styles.stat}>
+      <TableCell className={styles.label}>{ label }</TableCell>
+      <TableCell className={styles.value}>{ value }</TableCell>
```
k3ntako: might want to change `{ value }` to `{ value || "" }`.

Without that it gives this error: `checkPropTypes.js:19 Warning: Failed prop type: Invalid prop 'children' supplied to 'ThemedTableCell', expected a ReactNode.`

client/pages/Person/SeatingForm.js
```diff
@@ -70,6 +85,12 @@ class SeatingForm extends Component {
           source={PREFERRED_SIDES}
           onChange={this.onPreferredSide}
           value={this.state.preferred_side || ''} />
+        <Dropdown
+          auto
+          label="Sex"
+          onChange={this.onSex}
+          source={SEXES}
+          value={this.state.sex || ''} />
```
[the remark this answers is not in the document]
mienaikoe: Yea I noticed that too, but for like every dropdown...

I think it's a react-toolbox problem...

client/pages/Person/PersonPage.js
mienaikoe: Oof. I'll put birthday on its own line.

client/pages/Person/PersonPage.js
mienaikoe: true

client/pages/Person/SeatingForm.js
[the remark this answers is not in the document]
mienaikoe: https://github.com/react-toolbox/react-toolbox/issues/1870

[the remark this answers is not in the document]
mienaikoe: solved it by updating the key every time. According to the issue, the fix has a PR, and in the mean time, you have to force a re-render by changing the key with the newest epoch timestamp.

PR #57, opened 2019-03-28
Redux store visitor pattern

k3ntako:


client/models/Fetcher.js
```diff
@@ -35,7 +38,7 @@ export default {
     .catch(err => {
       //response handler doesn't catch following
       if(err.message === "Failed to fetch"){
-        store.dispatch(endServerCheck(false));
+        reduxStore.dispatch(endServerCheck(false));
```
mienaikoe: instead of doing this, can you export something from fetcher for workbox to call on an error?

The whole purpose of this ticket is separation of concerns: to remove dependencies on other parts of the project that are not this file's concern. In this case, workbox is something that should depend on Fetcher, because it's job is to know when/how to handle offline data and states. Fetcher's sole job is to wrap fetch in a nice api that returns json. Part of its job can be to create an event that notifies about an error in doing that job, but part of its job cannot be to tell workbox what's happening. It's up to workbox to listen for that as part of its job.

client/utilities/Workbox.js
```diff
@@ -36,7 +40,7 @@ const fetchServer = () => {
     cache: 'no-cache'
   }).then(response => {
     if( response.ok ){
-      store.dispatch(endServerCheck(true));
+      reduxStore.dispatch(endServerCheck(true));
```
mienaikoe: This uses redux enough that it really could be a component that renders nothing. 

Check out this article.  https://kyleshevlin.com/renderless-components

client/models/Fetcher.js
k3ntako: I wanna make sure I'm understanding you. It's not good that Fetcher is that one that checks if the app is connected to the server. That is Workbox's job, not Fetcher's?

What do you think would be a better alternative? I could have something in Workbox that checks every x seconds if the server is up. It seems like a waste of resources, especially for something that might never happen.

mienaikoe: No. Think of it this way:

If you, Kentaro, were Fetcher, it would be unreasonable for you to know the needs of every other file in the project. What's more reasonable is to have every other file tell you what they need from you, so that when the time does come that, say an error is thrown, you know exactly who to tell and how to tell them. 

Fetcher should not import from workbox, because that is forcing Fetcher to know that workbox needs its services. This is the same reason that workbox shouldn't be importing redux. Instead, workbox should import Fetcher and call something that it exports to allow Workbox to ask Fetcher to notify it when there's a server error. This way Fetcher isn't required to know the inner workings of the rest of the files, and workbox just needs to know that Fetcher has its back whenever it detects that the server is down.

k3ntako: Export a function, and then call all of the callbacks.

Publish subscribe model.

client/utilities/Workbox.js
k3ntako: put it in `index.js`

mienaikoe: Yo check this article out too. React hooks can subscribe to update really easily.

https://reactjs.org/docs/hooks-overview.html

k3ntako: wouldn't hooks be more for `state`, instead of redux?

mienaikoe: If you scroll about halfway down, there's an example of using hooks with pub/sub. I guess hooks aren't super important, but it should make updating your state based on whether Fetcher ran into an Error easier.

client/components/NetworkStatus/Workbox.js
```diff
@@ -0,0 +1,93 @@
...
+const MAX_MILLISECONDS = 10*60*1000;
+
+class Workbox extends Component {
+  constructor(props){
+    super(props);
+    this.checkServerStatus = this.checkServerStatus.bind(this);
+  }
+
+  componentDidMount(){
+    Fetcher.addServerErrorHandler(this.checkServerStatus);
```
mienaikoe: So Beautiful!!!

client/components/NetworkStatus/Workbox.js
```diff
@@ -0,0 +1,93 @@
...
+
+  expBackoff(delay){
+    let { isOnline, isServerConnected } = this.props;
+    if(!isOnline || isServerConnected){
+      return;
+    }
+
+    this.fetchServer();
+
+    if(delay >= MAX_MILLISECONDS){
```
mienaikoe: You can do this with:
`Math.min( MAX_MILLISECONDS, delay * 2 )`

client/components/NetworkStatus/index.js
```diff
@@ -79,6 +82,7 @@ class NetworkStatus extends Component {
         onTimeout={this.hideSnackbars}
         timeout={6000}
       />
+    <Workbox />
```
mienaikoe: nit: Indent this to show its in the same line as the snackbars

client/components/NetworkStatus/index.js
```diff
@@ -79,6 +82,7 @@ class NetworkStatus extends Component {
         onTimeout={this.hideSnackbars}
```
mienaikoe: For the comment on this snackbar, should it reflect something a bit more general like "Can't connect to the server" or something? The person might not be offline if the server is down for some reason. Saying that they're offline when they clearly have internet might be misleading and confusing. Maybe a third snackbar for specifically the server being down would be fine too.

client/components/StandIn/index.css
```diff
@@ -0,0 +1,19 @@
...
+}
+
+.loading{
+  position: fixed;
+  height: 60px;
+  width: 60px;
+  top: 50%;
+  left: 50%;
+  margin-top: -30px;
+  margin-left: -30px;
```
mienaikoe: Pro css tactics. You're in the big leagues now.

client/components/StandIn/index.js
```diff
@@ -0,0 +1,46 @@
...
+const StandIn = (props) => {
+  let { title, message, icon, isOnline, isServerConnected } = props;
+
+  if(props.loading){
+    return <div className={styles.loading}>
+      <ProgressBar
+        type="circular"
+        multicolor
+        mode="indeterminate"
+        theme={{ circular: "circle" }}
```
mienaikoe: the string `"circle"` just works? You don't need to reference it from styles?

client/components/StandIn/index.js
```diff
@@ -0,0 +1,46 @@
...
+        />
+    </div>
+  }else if(props.handleOffline && (!isOnline || !isServerConnected)){
```
mienaikoe: This goes to the point above. Saying the person is offline when the server is down would confuse and frustrate them (It makes them feel unfairly blamed for a workflow breakdown. Good copy writing takes responsibility for problems that the program can identify as its own and gives an actionable step people so they don't feel powerless and uninformed). 

Have a different message for either state.

client/index.js
```diff
@@ -13,11 +13,16 @@ import Routes from './Routes';
...
+
 const store = configureStore();
 
+Fetcher.addServerErrorHandler(() => store.dispatch(endServerCheck(false)));
+
 class App extends Component {
 
-  render () {
+  componentDidMount(){
     if ('serviceWorker' in navigator) {
```
mienaikoe: Is there ever a chance of a race condition where serviceWorker won't be in navigator?

client/models/Fetcher.js
```diff
@@ -1,6 +1,7 @@
-import { store } from './../index';
-import { endServerCheck } from './../redux/modules/workbox';
-import { checkServerStatus } from './../utilities/Workbox';
+let callbacks = [];
```
mienaikoe: make it `const`

I'd also call this something a bit more descriptive in case you end up adding more pub/sub callback arrays.

client/pages/EventForm/EventForm.js
```diff
@@ -243,11 +243,12 @@ class EventForm extends Component {
       return<div className='page'>
         <Heading title={pageTitle} />
         <section>
-          <Error
-            errorTitle={"Could Not Load Form"}
-            message={message}
+          <StandIn
+            title={"Could Not Load Form"}
```
mienaikoe: doesn't need brackets if it's a string

client/index.js
k3ntako: If you're asking about the `if` statement, that is just checking if `serviceWorkers` are supported by the browser.
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/serviceWorker



However, I've noticed that the first time you load the website after clearing storage in Chrome (under "Application" tab in Dev Tools) Workbox isn't loaded. On the second load, the service worker is cached so Workbox works as expected. This shouldn't be a big issue for most people, since they probably will load the page more than once before being offline.

Not sure how reliable this is but they say it's normal? 
https://stackoverflow.com/questions/49897182/workbox-sw-runtime-caching-not-working-until-second-reload

Also, the way I used to register the service work is the way Google's documentation recommends, so all of this combined means it's okay?

mienaikoe: Ohhh I see. Didn't know serviceworker was native

PR #59, opened 2019-03-29
Erg Time

mienaikoe:
- Added a field for erg_time to Person
- Added the field for Person.update
- Showing the average erg time for a boat
- Small bug fix for attendee list


client/pages/Boat/BoatSidebar.js
```diff
@@ -52,6 +52,8 @@ export default class BoatSidebar extends Component {
           <div className='statistic'>{ balanceLR }</div>
           <h4> Forward/Aft </h4>
           <div className='statistic'>{ balanceFA }</div>
+          <h4> Average 500m Split </h4>
+          <div className='statistic'>{ boat.averageErgTimeString() }</div>
```
k3ntako: I'm wonder if it's important that coach can see how many people have recorded times. Would it help to have like "18/22 have recorded times" under the average time? Or do you think coach would know?

Or you think it would be nice to have a column on the roster for whether someone has an ERG time or not? Currently, it's hard to figure out who doesn't.

client/pages/Event/EventAttendance.js
```diff
@@ -25,11 +25,15 @@ export default class EventAttendance extends React.Component {
...
       const boatCells = event.boats.map( boat => {
-        const boatSeat = personBoats[boat._id];
-        if( boatSeat ){
-          boatTotal++;
+        let boatSeat;
+        if( personBoats ){
+          boatSeat = personBoats[boat._id];
+          if( boatSeat ){
+            boatTotal++;
+          }
```
k3ntako: An issue with not showing the coaches on the attendance sheet is that I can't see the boat totals for them. Not sure how much it matters for our team. I think it might for other teams.

client/components/MinutesInput/index.js
```diff
@@ -0,0 +1,25 @@
...
+import styles from './index.css';
+
+
+export default function(props){
+  const value = props.value || 0;
+  const minutes = Math.floor( value / 60 );
+  const seconds = value % 60;
+
+  return <div className={styles.root}>
+    <label>500m Split Time</label>
```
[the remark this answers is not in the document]
mienaikoe: I think it's because github doesn't have a good jsx parser

client/pages/Boat/BoatSidebar.js
```diff
@@ -52,6 +52,8 @@ export default class BoatSidebar extends Component {
           <div className='statistic'>{ balanceLR }</div>
           <h4> Forward/Aft </h4>
           <div className='statistic'>{ balanceFA }</div>
+          <h4> Average 500m Split </h4>
```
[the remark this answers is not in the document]
mienaikoe: see above

client/pages/Boat/BoatSidebar.js
mienaikoe: Ohh true. and weight too.

mienaikoe: I'll add a couple columns to the roster for coach only

client/pages/Event/EventAttendance.js
mienaikoe: This was just a bug fix. What you're requesting is a different feature. Put it on trello and we can discuss it there. Not here.

mienaikoe: This was just a bug fix. I hear what you're saying, but don't want to have this ticket spin out of control with new things.

mienaikoe: alright I'll do it.

client/pages/Boat/BoatSidebar.js
```diff
@@ -53,7 +54,7 @@ export default class BoatSidebar extends Component {
           <h4> Forward/Aft </h4>
           <div className='statistic'>{ balanceFA }</div>
           <h4> Average 500m Split </h4>
-          <div className='statistic'>{ boat.averageErgTimeString() }</div>
+          <div className='statistic'>{ DurationUtils.durationToString(boat.averageErgTime()) }</div>
```
k3ntako: You aren't using the `statistic` className. There's one more right above it (line 55).

client/components/Table/index.css
```diff
@@ -0,0 +1,4 @@
+.root{
+  overflow-x: auto;
```
k3ntako: This is much better, but I noticed that there's no scroll bar before you start scrolling. If the table cuts off between columns, which happens a lot, I wouldn't know that I can scroll.

Is there a way to make the scroll bar visible? Maybe this can be in a different PR.

mienaikoe: I'll add a fade to the left/right when there's more

mienaikoe:
^ should be round 2 lol forgot we already had a round 1

client/components/Table/index.js
```diff
@@ -4,12 +4,59 @@ import { Table as ToolboxTable } from 'react-toolbox/lib/table';
...
+      overlayClass: styles.leftEnd,
+      overlayNeeded: false,
+    }
+    this.onScroll = this.onScroll.bind(this);
+  }
+
+  componentDidMount(){
+    if( this._refs_scrollable.scrollWidth > this._refs_scrollable.offsetWidth ){
+      this.setState({
+        overlayNeeded: true
```
k3ntako: what if they resize the window? or they rotate their phone/tablet? If they make their window smaller, the user wouldn't know that the table is cut off. I want this to be a reliable indicator.

mienaikoe: *sigh. yea. I was waffling about that yesterday. I think I should just add it.

PR #62, opened 2019-04-04
Introducing 200m and 2000m erg times

mienaikoe:


client/models/Boat.js
```diff
@@ -175,18 +177,40 @@ class Boat extends Model {
...
       }
     })
-    if( n === 0 ){
-      return 0;
-    }
-    return Math.round(sum / n);
+
+    const ret = {};
+    counters.forEach( counter => {
+      if( counter.n == 0 ){
```
k3ntako: Would it be better to use `===` or `counter < 1`?  `0 == ""` returns `true`. 

Or does it not matter because we know that it starts with `0` and it only goes up?

mienaikoe: In this case, it doesn't really matter because counter.n is always guaranteed to be a number.

However, I like to be explicit about these sorts of things, so I'll make it a triple equal.

PR #63, opened 2019-04-05
Event Cancellation

mienaikoe:
- Adds an endpoint for cancelling an event
- Checks for Coach on update, cancel, and delete endpoints
- Inroduces async/await to all EventController methods
- Adds a getter for isEditDisabled for an event, meaning nobody can attend it or edit it.
    Combines a check for cancellation and whether the event has passed.
- Adds checks for isEditDisabled all over the event pages and boat pages.


client/models/Event.js
```diff
@@ -101,6 +107,10 @@ class Event extends Model {
     });
   }
 
+  get isEditDisabled(){
+    return this.isCancelled || this.datetime < new Date();
```
k3ntako: It's not great that you can change the line-up after the event is over. what if we changed this to midnight after the event and disabled assignment based on this.

But, I guess an event can be more than one day?

client/pages/Event/index.js
```diff
@@ -77,42 +78,55 @@ class EventPage extends Component {
     };
   }
 
+  toEdit(){
+    this.props.history.push(`/events/${this.state.event._id}/edit`);
```
k3ntako: weren't you saying it's better to use `<Link>` for SEO. I guess that doesn't really matter?

server/controllers/EventController.js
```diff
@@ -162,49 +178,80 @@ const EventController = {
...
   },
 
-  delete: function(req, res, next){
-    return Event.findOneAndDelete({
+  delete: async function(req, res, next){
+    try {
+      const team = await Team.findById(req.session.team_id);
+      if( !team.coaches.find(coachId => coachId.equals(req.session.person_id)) ){
+        throw new ForbiddenError("You must be a coach to access this endpoint");
+      }
```
k3ntako: I feel like we need this check a lot. You think we could benefit from making this a utility function?

or like have something like `req.session.isCoach`? would that be a security issue?

client/models/Event.js
```diff
@@ -91,6 +91,12 @@ class Event extends Model {
     });
   }
 
+  cancel(){
+    return Fetcher.put(`/api/events/${this._id}/cancel`).then( () => {
+      this.isCancelled = true;
```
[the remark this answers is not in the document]
mienaikoe: I thought it might be a bit confusing if people see an event cancelled and then it's not cancelled later. But perhaps that's up to the coach.

client/models/Event.js
mienaikoe: yea I was worried about a couple things:
1) We aren't handling timezone well. What if a team goes to a race in like Hawai'i but they set the time in their local time (say they're from the UK). Now their midnight is actually noon in Hawai'i. The fix for this would be to handle timezone correctly, but it's just easier to let the coach edit forever until we fix that.

2) Like you say, a multi-day arrangement. What if someone goes to camp and has boat alignments for every day in a single event (That would be an awful way to organize things, but I've seen people do stupid things like that). Now they can't edit things on days 2-7. The fix for this would be to break datetime into a start and finish field and stop editing at the finish, but once again, just easier to let them edit forever until that fix is added.

client/pages/Event/index.js
mienaikoe: Yea. I realized this component would be 10x simpler if it wasn't a Link, and that it would never have SEO bots looking at it because it's a coach-only component.

server/controllers/EventController.js
mienaikoe: Yea I was considering making that a "before_filter" kind of like how we require authentication on some routes in the routes file. I think it would be fine to add an isCoach to the session when you login/join. It's not like the user can change anything in their session.

mienaikoe: actually is_coach on login is pretty brilliant. Check it once and never check it again.

PR #64, opened 2019-04-05
Colors for Draft.js

k3ntako:
Not sure how I can test if it actually sends colorful emails.


mienaikoe:
You can actually test by sending the email

client/components/Draft-js/ColorControls.js
```diff
@@ -0,0 +1,119 @@
...
+    const { active, colorStyleMap, label } = this.props;
+    let style = {
+      backgroundColor: colorStyleMap[this.props.style].color,
+    };
+
+    if(active){
+      style.border = '2px solid black';
+    }
+
+    return <span className={styles.colorBox} style={style} onMouseDown={this.onToggle}></span>
```
mienaikoe: what's wrong with onClick?

client/components/Draft-js/ColorControls.js
```diff
@@ -0,0 +1,119 @@
...
+      const active = currentStyle.has(type.style);
+      if(active){
+        activeColor.color = this.props.colorStyleMap[type.style].color;
+      }
+
+      return <StyleButton
+        key={type.style}
+        active={active}
+        onToggle={this.props.onToggle}
+        style={type.style}
```
mienaikoe: would suggest that you name this prop `colorStyle` or something a bit more descriptive than just `style`

client/components/Draft-js/ColorControls.js
```diff
@@ -0,0 +1,119 @@
...
+      e.preventDefault();
+      this.props.onToggle(this.props.style);
+      this.props.toggleShow(false);
+    };
+  }
+
+  render() {
+    const { active, colorStyleMap, label } = this.props;
+    let style = {
+      backgroundColor: colorStyleMap[this.props.style].color,
```
mienaikoe: if color is the only member of the object from `colorStyleMap`, can't you just use a string?

client/components/Draft-js/Editor.js
```diff
@@ -1,37 +1,93 @@
...
+
+      this.observer = new MutationObserver(this.onResize.bind(this));
+      this.observer.observe(targetNode, observerOptions);
+    }
+  }
+
+  componentWillUnmount(){
+    if(this.props.scroll){
+      this.observer.disconnect();
+      window.onscroll = null;
```
mienaikoe: What if something else is using window.onscroll.

You should use `addEventListener('scroll'` instead of directly calling `window.onscroll`

client/components/Draft-js/EditorTextarea.css
```diff
@@ -8,10 +8,13 @@
...
-  min-height: 200px;
 }
 
-#styleControlsContainer{
+.textEditor div[class*="public-DraftEditor-content"]{
+  min-height: 400px;
+}
+
+.styleControlsContainer{
   width: calc(100% - 2rem);
```
mienaikoe: :|

instead of using calc everywhere, can you use `left` and `right` ?

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "server/index.js",
   "scripts": {
     "test": "NODE_ENV=test mocha \"server/tests/*.js\" --exit",
-    "dev": "concurrently --raw \"webpack --mode=development --watch\" \"NODE_ENV=development node server/index.js\"",
+    "dev": "concurrently --raw \"webpack --mode=development --watch\" \"SET NODE_ENV=development& node server/index.js\"",
```
mienaikoe: there should be a space between the t and the &

client/components/Draft-js/Editor.js
```diff
@@ -57,27 +148,40 @@ export default class TextEditor extends Component {
...
   }
 
   render() {
-    let {editorState, focused} = this.state
+    const { focused } = this.state;
+    const { editorState, parentStyles, colorDirection } = this.props
 
-    let focusedClass = focused ? ` ${styles.focused}` : "";
     let hasTextClass = "";
+    if(parentStyles.hasText){
```
mienaikoe: why does parentStyles have a mix of theme classnames and booleans that represent something else? You can follow the `theme` props pattern that react-toolbox has, and put all the boolean stuff in different props. If it's a map of classNames that's meant to convey a theme, it should be just that and only that.

client/components/Draft-js/EditorInput.css
```diff
@@ -25,7 +25,7 @@
   border-color: rgba(21,21,21, 0.12);
 }
 
-.textEditor.focused > div:first-of-type{
+.textEditor.focused div[class*="public-DraftEditor-content"]{
```
mienaikoe: what? Why???

package.json
mienaikoe: Talked about this. If you're using windows, don't ruin the dev environment for others. Just make another script for your own development.

client/components/Draft-js/ColorControls.js
k3ntako: The Draft.js editor loses focus with `onClick` because `mousedown` comes before `click`

client/components/Draft-js/EditorInput.css
k3ntako: This is here so the editor to set the `min-height`. 

The editor is inside main layers of `div`s, and clicking on the outer ones doesn't produce the expected outcome. Mainly, clicking on the bottom right of `div:first-of-type` (aka editor) brought the text cursor to the beginning of the editor instead of the end.

The alternative is `.textEditor.focused > div > div > div` (I think it was three `div`s deep). I thought this was a better way of doing it.

client/components/Draft-js/Editor.js
k3ntako: So yeah, that was a dumb way of doing things. I was seeing if it was an Event or Email.

I added a `prop` called `editorType` which will now tell the Editor if it's an Event or Email.

client/components/Draft-js/ColorControls.js
mienaikoe: Oh you want to apply the color to the highlighted portion. got it.

client/components/Draft-js/Editor.js
mienaikoe: Was gonna say good, but the editor shouldn't know the specifics of its callers. It should only expose options that it knows about which either event or email can implement.

client/components/Draft-js/ColorControls.js
```diff
@@ -0,0 +1,119 @@
+import React, {Component} from 'react';
+import FontIcon from 'react-toolbox/lib/font_icon';
+
+import styles from './StyleButton.css';
+
+var COLORS = [
+  { style: 'red' },
```
mienaikoe: why not just an array of strings?

client/components/Draft-js/Controls.js
```diff
@@ -0,0 +1,94 @@
+import React, {Component} from 'react';
+import {EditorState, Modifier, RichUtils } from 'draft-js';
+
+import BlockControls from './BlockControls';
+import InlineControls from './InlineControls';
+import ColorControls from './ColorControls';
+
+
+export default class TextEditor extends Component {
```
mienaikoe: your class is named `TextEditor` but it's in a file named `Controls`. I suggest naming this `FormatControls` or something like that.

client/components/Draft-js/Editor.js
```diff
@@ -1,16 +1,33 @@
...
+  },
+  blue: {
+    color: '#0288d1',
+  },
+  black: {
+    color: '#000',
+  }
+};
 
 export default class TextEditor extends Component {
```
mienaikoe: Again, class is named `TextEditor` but appears in a file named `Editor`

client/components/Draft-js/ColorControls.js
k3ntako: I'm a bit confused with what you mean. 

I think `colorStyleMap` is an object because that's what Draft.js expects for custom styles.

client/components/Draft-js/Editor.js
k3ntako: I changed it to: `hasLabel` and `colorPickerDirection`

client/components/Draft-js/ColorControls.js
mienaikoe: Ahhh ok.

client/components/Draft-js/EditorInput.css
mienaikoe: What's wrong with just setting min-height on `.textEditor`

I tried it locally and it seems to work just fine.

I would also add a `cursor: text`.

mienaikoe: https://github.com/facebook/draft-js/issues/528#issuecomment-262862767

PR #65, opened 2019-04-06
Boat Copy

mienaikoe:
- Adds an endpoint to copy a boat
- Permits copying a boat as another action in the ellipsis dropdown on Boat Page


server/routes.js
```diff
@@ -80,6 +80,8 @@ router.route( '/boats/:boat_id').
   get(    BoatController.find).
   put(    BoatController.update).
   delete( BoatController.delete);
+router.route( '/events/:event_id/boats/:boat_id/clone').
+  post(   BoatController.clone);
```
k3ntako: we could as `requireCoach` here

mienaikoe: Yep. Was waiting on the PR that had requireCoach to go out before added that

PR #66, opened 2019-04-08
Remove sensitive info from API responses

k3ntako:
Sorry about all the spaces, Atom deletes those automatically.
- Hid sensitive info (erg times, weight, sex, birthday) from members
  - Hid the info from coach if coach didn't need it.
- if birthday is needed, the year was set to `0` in controller before returning response
- added `requireCoach` in a few places.

Should we remove shirt sizes from the roster?
In the roster tabled, you used a check mark for CPR Certified. Should that be a `<FontIcon>`. I thought you mentioned that not every device can read that.


client/pages/Person/PersonPage.js
```diff
@@ -70,17 +70,24 @@ class PersonPage extends Component {
...
             <Table selectable={false}>
               { this.renderStat("Email", person.email) }
               { this.renderStat("Phone", person.phone) }
-              { this.renderStat("Birthday", EventFormatter.formatBirthday(person.birthday)) }
+              { this.renderStat("Birthday", birthday) }
               { this.renderStat("Neighborhood", person.location) }
-              { this.renderStat("Shirt Size", person.shirt) }
-              { this.renderStat("CPR Certified", person.cpr_cerified) }
+              { this.renderStat("Shirt Size", person.shirt_size) }
+              { this.renderStat("CPR Certified", person.cpr_certified ? '✔' : '') }
```
k3ntako: change to FontIcon

server/controllers/PersonController.js
```diff
@@ -74,7 +81,7 @@ const PersonController = {
   me: function(req, res, next){
     return Person.
       findOne({ _id: req.session.person_id }).
-      select(SAFE_SELECT).
+      select(BIRTHDAY_SELECT).
       populate(DEEP_POPULATOR).
```
k3ntako: we don't need invited, members

we need coaches for isCoach();

PR #67, opened 2019-04-08
Added "required" to client models and bug fixes

k3ntako:
- Front-end models were missing `required`.
- Add `required` to fields in `EventSeriesSchema` on the back-end
- `Model` had a bug where if the field was `false` it would set the field as `null`.
- Fixed bug where "Show/Hide Boat" text wouldn't toggle.


client/models/Assignment.js
```diff
@@ -45,6 +45,7 @@ class Assignment extends Model {
 Model.registerFields(Assignment, 'Assignment', [
   {
     key: "name",
+    required: true
```
mienaikoe: `required` is only used on backend models (Mongoose)

mienaikoe:
lgtm

PR #69, opened 2019-04-09
Change Email Addresses

mienaikoe:
- Modify mskdbadministrator@gmail.com to be admin@wakebird.com 
- Change sending email from `noreply` to `hello` (I suspect gmail might flag noreply email addresses, or emails that don't have a record to send to)
- Fixing issues with Model.js from removing certain fields.

I've added mx records to our domain so that replies to hello and admin@wakebird.com both go to the Mighty Speed Krush gmail account


server/utilities/EmailUtils.js
```diff
@@ -15,7 +15,7 @@ const MAIL_SETTINGS = {
 
 const DEFAULT_SENDER = {
   name: 'Wakebird',
-  email: 'noreply@wakebird.com'
+  email: 'hello@wakebird.com'
```
k3ntako: Should we make this "noreply"? We don't want people email this?

mienaikoe: 1) hello@wakebird.com is now forwarded to the email account you made, so replies are ok.

2) I also read that noreply email addresses might be marked as not the greatest for spam reasons.

k3ntako: number 2 is especially important!

PR #70, opened 2019-04-09
Notification Settings

mienaikoe:
- Creates a backend endpoint for updating settings
- Creates a form on edit account for updating settings
- Changes the reminder task to listen to the notification setting


client/pages/Me/Settings.js
```diff
@@ -0,0 +1,38 @@
...
+    setNotifications( !notifications )
+    setSaved(false)
+  }
+
+  function save(){
+    props.me.updateSettings( notifications ).then( () => {
+      props.issueSnackbar("Your settings have been saved.")
+      setSaved(true)
+    })
+  }
```
k3ntako: question: are these functions (`save()` and `toggle()`) being redefined every time `Settings` is mounted? 

If so, is it more efficient to have a class and instance methods?

server/models/Person.js
```diff
@@ -71,7 +71,10 @@ const PersonSchema = mongoose.Schema({
     reset_token:      { type: String, default: null },
     token_expiration: { type: Date, default: null },
   },
-  team:               { type: ObjectId, ref: 'Team', index: true }
+  team:               { type: ObjectId, ref: 'Team', index: true },
+  settings: {
+    attendance_notifications: { type: Boolean, default: true, requiired: true }
```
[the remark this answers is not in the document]
mienaikoe: :O

client/pages/Me/Settings.js
mienaikoe: hmm I'm not sure what react is doing behind the scenes. It might construct a component from the signature of this function... or something. Either way, this is the way react instructs us to use hooks.

mienaikoe: An interesting read. Check the comments too to get an alternative view on the issue:

https://medium.freecodecamp.org/7-reasons-to-outlaw-reacts-functional-components-ff5b5ae09b7c

k3ntako: that brought up a lot of things I had never consider, but it seems like people can't 100% agree on what's best. i think we can keep like this?

I discovered `React.PureComponent` and `React.memo` through reading this article. I wonder if we can start using them moving forward. Have you used them in the past?

mienaikoe: Ooh. I've used `PureComponent` a long time ago, but honestly forgot about it. I think the react devs basically went full blast on functional components and didn't say much about purecomponents after like 2017

Never heard of `React.memo`.

PR #71, opened 2019-04-11
Verify email

k3ntako:
Two paths:
- Sends invite email with a join link that will verify as user creates an account.
- After creating an account, if the user is not verified from path above, an email will be sent to the user.
  - User can have the email resent. Similar to the `EventCreateCard` for coach on the homepage, there will be a big banner asking them to verify their email. 

`!this.team || !this.team._id `
- Seems like the changes in `Model` now create an empty object when `Person.team` is falsy, so we need to check if the `id` exists.

Test: Didn't write tests for requesting a verification email. There's no easy way to know how to know the token that was sent out.

Existing users will not be verified. We will verify coaches through a migration.

To do:
- Welcome email for verified users. (Future PR)
- ~~Disable buttons for non-verified users.~~ (DONE)
- ~~Write tests~~ (DONE)
- Coach should see who hasn't verified their email (Future PR)
- ~~write query string parser~~ (Used "qs" npm package)


k3ntako:
I will do the last two todos in a future PR. will make Trello cards.

client/pages/Team/VerifyEmailCard.css
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.verifyRootMobile{
+  display: none;
+}
+
+.verifyRoot .title{
+  font-family: 'Oswald' !important;
+  font-weight: 300 !important;
+  font-size: 3.5rem !important;
```
mienaikoe: if using CardTitle requires you to use !important everywhere, I just wouldn't use it because you're basically defying all of the styles that CardTitle gives you out of the box. You can just use a div.

server/controllers/EventController.js
```diff
@@ -5,7 +5,8 @@ const Event = require('../models/Event')
 const Team = require('../models/Team')
 const Assignment = require('../models/Assignment')
 
-const SAFE_SELECT = "-auth -weight -preferred_side -preferred_position -birthday -erg_times -sex";
+const UNVERIFIED =  " -email -location -phone -shirt_size -cpr_certified "
```
mienaikoe: would call this `UNVERIFIED_SELECT` so it's clear what it's to be used for.

server/controllers/PersonController.js
```diff
@@ -1,12 +1,17 @@
 const { ValidationError, ForbiddenError, NotFoundError } = require('../errors');
+const crypto = require('crypto');
+
 const Person = require('../models/Person')
 const Team = require('../models/Team');
-const EmailUtils = require('../utilities/EmailUtils')
+const { passhashify, generateEmailToken } = require('../utilities/AuthUtils')
+const EmailUtils = require('../utilities/EmailUtils');
 
+const ONLY_VERIFIED = " -verification.verification_hash -verification.salt -verification.token_expiration "
```
mienaikoe: same here. Would add `SELECT` to the end.

mienaikoe: Also, why not make all these things separate from the `is_verified` flag? They technically are separation of concerns, because all of these fields involve the actual process of verification while the `is_verified` flag controls what features you get to use on Wakebird.

server/controllers/PersonController.js
```diff
@@ -206,17 +220,32 @@ const PersonController = {
         new: true // https://stackoverflow.com/questions/35626040/findoneandupdate-used-with-returnnewdocumenttrue-returns-the-original-document
       });
 
+      const expire = person.verification.token_expiration;
+      let joinLink = "";
+
       if( person.auth.passhash ){
         team.members.push(person);
+      } else if( !expire || expire.getTime() < Date.now() ) {
```
mienaikoe: you can simplify this by keeping the original if/else, and just wrapping this joinLink work inside of the else block.

server/controllers/SessionController.js
```diff
@@ -4,10 +4,12 @@ const crypto = require('crypto');
 const { ValidationError, ForbiddenError, NotFoundError } = require('../errors');
 const Person = require('../models/Person')
 const Team = require('../models/Team')
-const passhashify = require('../utilities/passhashify')
+const { passhashify, generateEmailToken, verifyEmailToken } = require('../utilities/AuthUtils')
```
mienaikoe: nice! this is smart.

server/controllers/SessionController.js
```diff
@@ -4,10 +4,12 @@ const crypto = require('crypto');
...
 const EmailUtils = require('../utilities/EmailUtils')
 
-const SAFE_SELECT = "-auth -weight -preferred_side -preferred_position -birthday -erg_times -sex";
+const ONLY_VERIFIED = " -verification.verification_hash -verification.salt -verification.token_expiration "
```
mienaikoe: Add SELECT to the end

server/controllers/SessionController.js
```diff
@@ -111,6 +123,7 @@ const SessionController = {
       }
 
       req.session.person_id = person._id
+      req.session.verified = person.verification.verified;
```
mienaikoe: would call this `is_verified` to clarify that it's a boolean, and to match the pattern of `is_coach`

server/controllers/SessionController.js
```diff
@@ -279,6 +293,82 @@ const SessionController = {
...
+        req.session.verified = person.verification.verified;
+      }
+
+      res.json({okay: true});
+      next();
+    } catch(err) {
+      console.error(err);
+      next(err);
+    }
+  },
```
mienaikoe: This is really beautifully written code. Your writing has improved a lot.

server/emails/join_with_verification/html.pug
```diff
@@ -0,0 +1,19 @@
+doctype html
+head
+body
+  p Hello #{first_name},
+  p
+    div Thanks for signing up on Wakebird!
+
+  p
+    div Please use this link to verify your email.
+    Certain features will be unavailable
```
mienaikoe: Did you verify that multiline text like this works? It seems odd for pug to work this way.

server/emails/resend_verification/text.pug
```diff
@@ -0,0 +1,11 @@
+= `Hello ${first_name},`
+
+= "Thanks for signing up on Wakebird!"
+
+= "Use this link to verify your email. Please verify your email to unlock all features on Wakebird."
```
mienaikoe: The text in this is different than in the html version

server/migrations/3_TestCoachEmail.js
```diff
@@ -2,25 +2,76 @@ const mongoose = require('mongoose');
...
+
+    console.log(wakebirdAdmin);
+  }
+
+  //make sure admin@wakebird.com is a coach
+  let msk = await Team.findMSK();
+  if( !msk.coaches.includes(wakebirdAdmin._id) ){
+    msk.coaches.push(wakebirdAdmin._id);
+    msk = msk.save();
+  }
```
mienaikoe: thanks for doing this.

server/migrations/3_TestCoachEmail.js
```diff
@@ -2,25 +2,76 @@ const mongoose = require('mongoose');
...
+      seats: {
+        $elemMatch : { person: person._id }
+      },
+    },{
+      $pull: {
+        seats: {
+          $elemMatch : { person: person._id }
+        },
+      }
+    })
```
mienaikoe: damn i didn't realize how involved this would be.

server/routes.js
```diff
@@ -32,112 +32,123 @@ function requireCoach(req, res, next){
   }
 }
 
+function requireVerified(req, res, next){
```
mienaikoe: I imagine `requireCoach` should probably require a verified state for basically any action. It might not be necessary for our team, but later on, it would.

server/tests/SessionControllerTest.js
```diff
@@ -235,4 +234,36 @@ describe('SessionController', function() {
...
+        },
+      }
+      const res = {
+        json: function(response){
+          assert(response.okay)
+
+          Person.findOne({ email: "admin@wakebird.com" }).exec().then( person => {
+            assert(person.verification);
+            expect(person.verification.verified).to.equal(false);
+            assert(person.verification.verification_hash);
```
mienaikoe: you don't really have to assert all these things if you're also checking for value or type. `expect` will handle assertion problems for you and 9/10 will give you better error messages for the problem.

server/utilities/AuthUtils.js
```diff
@@ -0,0 +1,66 @@
...
+      }
+      return person;
+    }
+
+    const calculatedHash = passhashify(verification_token, verification.salt);
+    if( calculatedHash === verification.verification_hash ){
+      person.verification.verified = true;
+      person.verification.verification_hash = null;
+      person.verification.salt = null;
+      person.verification.token_expiration = null;
```
mienaikoe: do all this on the controller side. Utilities shouldn't modify documents.

Additionally, you need to save. If you are saving, that's another reason to do this on controller side, so we don't have to go back and forth between files to see where it gets saved.

client/models/Person.js
```diff
@@ -113,7 +127,7 @@ class Person extends Model {
   }
 
   isCoach(){
-    if( !this.team ){
+    if( !this.team || !this.team._id ){
```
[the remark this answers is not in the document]
k3ntako: Not sure anymore. it was causing an issue, but can't recreate it.

client/pages/Auth/ResetPassword/index.js
```diff
@@ -39,19 +39,6 @@ export default class ResetPassword extends React.Component {
...
-      return <div>
-        <p className={styles.error}>
-          { this.state.error }
-        </p>
-        <p>
-          <Link to="/send-reset">Please try again</Link>
-        </p>
-      </div>
-    }
-  }
```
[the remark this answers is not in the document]
k3ntako: this wasn't being used

client/pages/Auth/VerifyEmail/index.js
```diff
@@ -0,0 +1,126 @@
...
+      error: null,
+      loading: true,
+      verified: false
+    };
+  }
+
+  componentDidMount(){
+    let queryString = this.props.location.search;
+
+    if(!queryString.startsWith("?vt=")){
```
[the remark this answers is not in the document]
k3ntako: i'm really not sure why I didn't use qs here.

server/controllers/SessionController.js
k3ntako: aww 😃 it's all thanks to you!

server/routes.js
k3ntako: I'm a little confused by this comment.

client/pages/Auth/ResetPassword/index.js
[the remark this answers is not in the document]
mienaikoe: true

server/routes.js
mienaikoe: The comment isn't specifically about this function. I was just thinking that the `requireCoach` function should call `requireVerified` as a first step. Any function with `requireCoach` should never allow an unverified account to make changes.

server/controllers/PersonController.js
k3ntako: and change to `is_verified`

PR #72, opened 2019-04-12
Raleway

mienaikoe:
- Fixed issues with Boat Mass and Series Form
- Replaced Comfortaa with Raleway


client/components/Nav/index.css
```diff
@@ -21,8 +21,8 @@
 }
 .logo::after{
   content: 'wakebird';
-  font-family: comfortaa;
-  font-weight: bold;
+  font-family: 'Raleway';
```
k3ntako: can we keep the logo in Comfortaa? here and the Login page.

client/pages/Team/EventCard.css
```diff
@@ -1,50 +1,49 @@
...
   align-items: flex-start;
-  padding: 0 2rem;
-}
-
-.cardTitle{
-  font-family: 'Oswald', sans-serif;
-  font-size: 28px !important;
+  padding: 2rem;
+  font-weight: 500;
+  color: var(--palette-grey-700);
```
k3ntako: I noticed that you changed Oswald to Raleway on the homepage. It feels less aggressive, which I like, but I feel like I'm not sure what info is supposed to grab my attention.

What if we changed the title of the event to Oswald?

public/index.css
```diff
@@ -12,12 +12,12 @@ body{
   margin: 0;
   font-size: 12px;
   background-color: var(--palette-blue-grey-50);
```
k3ntako: i feel like our website looks so grim? what if we changed the colors of the headers?

client/pages/Team/EventCard.css
mienaikoe: mm since they're all the same font size I think that would be a bad idea - it would cause an even bigger clash of message. 

We had this same issue when it was Oswald - the date, time, and name are all the same size.

Since the only things that change are the date and time, I think they should be more prominent, but if it's a race, we want the name to be prominent. Should this just be a set of calendars?

public/index.css
mienaikoe: It's a bit grim, yea. but I think it's caus we got used to a really bright blue.

What would you want the headers to be?

client/pages/Team/EventCard.css
mienaikoe: Look at TeamSnap.
1) Group by Month
2) Date on the left side
3) Time in addition to Location
4) Increase size of time & location by a little bit

public/index.css
mienaikoe: Let's try to make the headers blue-800 ish

mienaikoe: tried em.. think we need to sit on it for a bit and think.

Maybe a new PR for the heading colors

client/pages/Auth/index.css
```diff
@@ -40,19 +42,12 @@
...
-}
-.logoTitle{
-  font-size: 2.7rem;
-  line-height: 5rem;
-  font-weight: bold;
-  text-transform: lowercase;
-  letter-spacing: 0.1em;
+  height: 140px;
+  margin: 0 auto 3rem;
+  background-image: url('/logo-splash-dark@600.png');
```
k3ntako: this image looks less smooth compared to the old one. could we increase the resolution?

client/pages/Auth/index.css
```diff
@@ -15,7 +17,7 @@
 }
 
 .authOverlay{
-  background: linear-gradient(to top right, var(--palette-blue-900), rgba(0, 100, 250, 0.5));
+  background: linear-gradient(to top right, rgba(0, 102, 255, 0.7), rgba(0, 191, 255, 0.7));
```
k3ntako: i feel like the new colors make the picture look a little more washed out and grainy. maybe the darker colors helped hide the imperfections.

PR #74, opened 2019-04-14
Email fix

k3ntako:
- `p` tags in our emails do not have margin. ONLY for team messages (not automated emails like join and reminder emails)
- Fixed issue where emails were not being sent with color.
- Draft editor: background of formatting buttons are only gray in edit event
- Draft editor: fixed issue where formatting buttons didn't fit on smaller devices. 
- Draft editor: email editor (not event note editor) is in `sans-serif` to match what recipient will see.

Existing bug:
- On mobile, formatting buttons retain `:hover` styles after user taps on it. 
  - First image: Underline was just selected, but the `:hover` effect is active.
  - Second image: Bold was unselected, but the `:hover` effect is active.
Screen Shot 2019-04-14 at 12 02 19 PM
Screen Shot 2019-04-14 at 12 02 28 PM


client/components/Draft-js/index.js
```diff
@@ -73,7 +73,7 @@ export default class TextEditor extends Component {
     const editor = document.getElementsByClassName(parentStyles.styleControlsContainer)[0];
     const windowHeight = window.innerHeight;
     const windowWidth = window.innerWidth;
-    const mobileAdjust = windowWidth <= 600 ? 54 : 0;
+    const mobileAdjust = windowWidth <= 700 ? 54 : 0;
```
k3ntako: move save, delete and send to bottom on desktop

mienaikoe: yeh. not much reason to float this around, but take a look and make a judgment call before just committing.

client/components/DraftReader/Reader.css
```diff
@@ -1,3 +1,20 @@
 .textReader > div{
   line-height: 2rem;
 }
+
+.textReader.sansSerif > div{
+  font-family: sans-serif;
+}
+
+.textReader.sansSerif > div h1, h2 {
```
mienaikoe: this should be
```
.textReader.sansSerif > div h1, 
.textReader.sansSerif > div h2
```

mienaikoe: (because we aren't using less or sass yet)

client/components/EmailTextEditor/EmailEditor.css
```diff
@@ -3,6 +3,19 @@
   font-family: sans-serif;
 }
 
+.textEditorRoot h1, h2 {
```
mienaikoe:
```
.textEditorRoot h1, 
.textEditorRoot h2
```

server/controllers/EmailController.js
```diff
@@ -41,6 +41,29 @@ const DEEP_POPULATOR = [
 
 const UPDATE_FIELDS = ["subject", "emailBody"];
 
+const INLINE_STYLES = {
```
mienaikoe: Can you put these styles into a special custom email utility? Or at least somewhere separate from the controller. I think this styling logic and config is crowding the main purpose of this controller.

PR #75, opened 2019-04-15
Email verification bug fixes

k3ntako:


client/pages/Auth/VerifyEmail/index.js
```diff
@@ -68,7 +79,8 @@ class VerifyEmail extends Component {
   }
 
   redirectToHome(){
-    this.props.history.replace('/');
+    // not react-router because props.me has to update
+    window.location.href = '/';
```
mienaikoe: MeContext exposes meState which has a function called `setMe` you can call to update the value of `me`. It was designed specifically for this case. After calling setMe, you can use history.replace.

client/pages/Auth/VerifyEmail/index.js
```diff
@@ -87,12 +99,10 @@ class VerifyEmail extends Component {
 
   render(){
     const { error, loading, isVerified } = this.state;
-    let inside;
-    if(loading){
-      inside = <div className={styles.verifyMessage}><h2>Verifying your email...</h2></div>
-    } else if (isVerified || error && error.includes("already verified")) {
+    let inside = <div className={styles.verifyMessage}><h2>Verifying your email...</h2></div>;
+    if (isVerified || error && error.includes("already verified")) {
```
mienaikoe: could you set the error to be an object with a message and some enumerated code instead? Using `includes` to determine the state of a request is not great.

client/pages/Team/Roster.js
```diff
@@ -131,7 +131,7 @@ class Roster extends React.Component {
       headCells.push(<TableCell key="erg_time_2000">2000m Split</TableCell>)
     }
 
-    return <Card>
+    return <Card key={"members_table"}>
```
mienaikoe: string attributes in react don't need to be in brackets

client/pages/Auth/VerifyEmail/index.js
k3ntako: so you mean a status code?

currently, Fetcher doesn't return the status of the error. it only returns a string (the error message). I think we should change that.

If we do, then we could send a 409. How does that sound?

> The HTTP 409 Conflict response status code indicates a request conflict with current state of the server.
>
> Conflicts are most likely to occur in response to a PUT request. For example, you may get a 409 response when uploading a file which is older than the one already on the server resulting in a version control conflict.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409

mienaikoe: You can send any sort of error code you want in the response body. Considering there are 3 different states verify could be in, and you want to handle them differently, I would suggest doing that instead of an HTTP status code.

Login and Join will need this too eventually. Let's use verify as a model for how to handle the status correctly.

k3ntako: So instead of throwing an error in the controller, I could `res.json({ okay: false, status: 475 })`. 475 as an example of a code that's not already taken.

On the front end we can have an object with all the status codes and their messages.
```
const STATUS_CODES = {
  475: "User is already verified",
  476: "...",
  ....
}
```

mienaikoe: No. Don't Use HTTP status codes. Get that out of your mind.

Create an enumerated set of values for status. Maybe the values can be strings like "already_verified" or something.

client/contexts/SnackbarContext.css
```diff
@@ -0,0 +1,7 @@
+.success{
+  background-color: var(--palette-green-600);
+}
+
+.error{
+  background-color: var(--palette-red-600);
```
k3ntako: looks like we are using `-500` for errors. it'd be better to standardize?

mienaikoe: yea. Always better to standardize.

Maybe you can add a variable to :root? I'm not familiar with how that works, but react-toolbox does that,

client/pages/Auth/VerifyEmail/index.js
k3ntako: `Statuses.js` takes in a status code like `"already_verified"` and then returns an object. 

Colors snackbar poses some new issues. 
- If we have three possible states success (green), error (red), and neutral (black), how do we represent that using one field? I chose to use two fields `success` and `error`.

does this look something like what you were thinking of?

k3ntako: `Statuses.js` takes in a status code like `"already_verified"` and then returns an object. 

Colors snackbar poses some new issues. 
- If we have three possible states success (green), error (red), and neutral (black), how do we represent that using one field? I chose to use two fields `success` and `error`.

does this look something like what you were thinking of?

mienaikoe: I think you tried to bake everything into the status code, which also isn't a great approach. Let's get on the phone about this one tomorrow.

k3ntako: return 400 with new error code.

Don't return in AuthUtils, catch it in controller and send it back

server/errors.js
```diff
@@ -8,7 +8,16 @@ class ForbiddenError extends Error {
 class NotFoundError extends Error {
   get code(){ return 404; }
 }
+class CustomError extends Error {
```
mienaikoe: Really hate to do this, but `CustomError` is a really generic and non-descriptive name for a class. Literally anything that overrides a class is "custom". Pick something that describes what its purpose is, like CodedError or BadRequestError.

PR #76, opened 2019-04-16
New Team Page

mienaikoe:
- Introducing a new Event Card Design, with calendar squares on the left
- Simplified CSS Event Card
- Introducing an `InspirationCard` Component that changes with the state of the user
- Adding a card for updating details if you haven't set your birthday or phone number, both of which are important for people on the team to contact you.
- Added common styles for different cards.
- Minor bug fixes.


client/pages/Team/EventCard.js
```diff
@@ -69,19 +31,19 @@ class EventCard extends Component {
...
-          <span> {event.attendees.length} </span>
+          <span className={styles.number}> {event.attendees.length} </span>
           <label> Attending </label>
         </div>
         <div key="responses">
-          <span> {event.attendees.length + event.non_attendees.length} </span>
+          <span className={styles.number}> {event.attendees.length + event.non_attendees.length} </span>
           <label> Responded </label>
         </div>
       </div>
```
k3ntako: on the iPhone SE (width < 350px) I think we could consider hiding these numbers. but, neither coaches have an SE sized phone.

client/pages/Team/EventCard.js
```diff
@@ -117,18 +79,40 @@ class EventCard extends Component {
...
+    const eventName = event.name + (event.is_cancelled ? ' [CANCELLED]' : '')
+    let cardTitle = <div className={styles.main}>
+      { eventName }
     </div>
-  }
+    let cardAction;
+    if( me.isCoach() || !event.is_cancelled ){
+      cardTitle = <Link to={`/events/${event._id}`}>
+        { cardTitle }
+      </Link>
```
k3ntako: Especially on mobile it would be nice if we could click on the card. 
I'm guessing the issue is the attendance buttons? 

I feel like not many people would use the right side to go to an event. And if they are trying to respond and they accidentally go to the event, it's not a big deal? They can respond there and go back?

client/pages/Team/InspirationCard/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+function InspirationCard(props){
+  
+  const { me, events } = props;
+
+  if( !me ){
+    return null;
+  }
+
+  if( !props.isOnline ){
```
k3ntako: maybe check if user is connected to server too?

server/migrations/3_TestCoachEmail.js
```diff
@@ -10,14 +10,16 @@ module.exports = async function(){
...
-    console.log("deleted: ", deleted.email);
-  }else{
+  if(wakebirdAdmin){  
+    if( dbAdmin ){
+      let updated = await removeFromTeam(dbAdmin);
+      console.log("boats: ", updated);
+      let deleted = await Person.findOneAndDelete({ email: 'mskdbadministrator@gmail.com' });
+      console.log("deleted: ", deleted.email);
+    }
+  } else if( dbAdmin ){
```
k3ntako: we check for `dbAdmin` twice. Wouldn't it be cleaner if we checked for `dbAdmin` and then checked for `wakebirdAdmin` inside?

client/pages/Team/EventsTab.js
```diff
@@ -1,85 +1,36 @@
...
 
-  toggleCreateDialog(){
-    this.setState( prevState => {
-      return {
-        createDialog: !prevState.createDialog
-      };
+  if( events && events.length ){
+    const eventsHtml = events.map(eventGroup => {
+      const monthName = EventFormatter.MONTH_NAMES[eventGroup[0].datetime.getMonth()]
+      return <div key={monthName}>
```
k3ntako: potentially could be an issue if user has events in the same month in two different years.

client/pages/Team/InspirationCard/EditMeCard.css
```diff
@@ -0,0 +1,15 @@
+.card{
+  background: url('/images/huddle.jpg');
+  background-position: 0 -220px;
+}
+
+.card .cardTitle::after{
+  content: "➤";
```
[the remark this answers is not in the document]
mienaikoe: css content is ok if it's not actual writing. This is an icon that shows on hover, that shows that something is a link, which I would argue is an ok use of this.

client/pages/Team/EventCard.js
mienaikoe: oof. Yeh I think 5 SE and below has like <5% market share

mienaikoe: you know, it would be easy to do that tho, so i'll just do it.

client/pages/Team/EventCard.js
mienaikoe: Yea the issue is the attendance buttons.

I think having to click on the title to visit the event is a lesser cognitive hurdle than accidentally clicking on the card and having to go back.

server/migrations/3_TestCoachEmail.js
mienaikoe: no because you'd still need another if statement for if there's neither wakebirdAdmin or dbAdmin.

In most cases of logic, you can't reduce the number of if/else statements, you can only rearrange them.

client/pages/Team/EventsTab.js
mienaikoe: I was considering windowing the upcoming events to within the next 12 months, and windowing past events to within the past 12 months. Let's cross that bridge when we get to it.

client/pages/Team/EventCard.js
k3ntako: I feel like people wouldn't accidentally tap the card very often, and i think it's not super clear that only the event name is clickable at first.

server/migrations/3_TestCoachEmail.js
k3ntako: Couldn't we do something like this? We actually aren't doing anything when there is no `dbAdmin`. Maybe that's an issue?

```
let wakebirdAdmin = await Person.findOne({ email: 'admin@wakebird.com' });
let dbAdmin = await Person.findOne({ email: 'mskdbadministrator@gmail.com' });

if(dbAdmin && wakebirdAdmin){
  let updated = await removeFromTeam(dbAdmin);
  console.log("boats: ", updated);
  let deleted = await Person.findOneAndDelete({ email: 'mskdbadministrator@gmail.com' });
  console.log("deleted: ", deleted.email);
} else if (dbAdmin) {
  wakebirdAdmin = await Person.findOneAndUpdate({
    email: 'mskdbadministrator@gmail.com',
  }, {
    email: 'admin@wakebird.com',
    first_name: 'Wakebird',
    last_name: 'Admin'
  }, {
    new: true
  });
}

```

mienaikoe: Still uses three if checks.

I see a way this would be more readable at least, so i'll try that.

client/pages/Team/EventCard.js
mienaikoe: I just disagree.

The event name is in the link color to make it clear that it's clickable. People might also expect clicking on the date or place or time does something specific to the date, place, or time, but it's in a gray unclickable color for a reason.

This is how the event card was before. I didn't change that. Let's get feedback before making any changes to this card.

PR #78, opened 2019-04-18
Email invites

k3ntako:


client/pages/Email/Recipients.js
```diff
@@ -169,6 +196,7 @@ class Recipients extends Component {
             <TableCell> </TableCell>
           </TableHead>
           {this.renderRecipients(this.props.email.nonMember, "face")}
+          {this.renderRecipients(this.state.excludedInvites, "highlight_off")}
           {this.renderRecipients(this.state.excludedMembers, "highlight_off")}
           {this.renderRecipients(this.state.includedMembers, "check_circle_outline")}
```
mienaikoe: This seems sorted in reverse order than I would intuitively think to sort it.

server/controllers/EmailController.js
```diff
@@ -73,18 +60,37 @@ const EventController = {
...
+        }else{
+          invites.push(person);
         }
+      })
 
-        res.json(emailMongoose);
-        next();
-      }).
-      catch( next )
+      invites = invites.sort(EmailUtils.sortByEmail);
```
mienaikoe: don't put one-liners into a utility object. This can easily be expressed as

`invites.sort( (a,b) => a.email > b.email )`

server/migrations/5_AddAdminToMSK.js
```diff
@@ -0,0 +1,30 @@
+const mongoose = require('mongoose');
+
+const Person = require('../models/Person');
+const Team = require('../models/Team');
+
+
+module.exports = async function(){
+  try {
+    console.log('========= 5_AddAdminToMSK');
```
mienaikoe: This should already be part of migrations 2_ and 3_

If those migrations are flawed, you should fix those migrations.

server/utilities/EmailUtils.js
```diff
@@ -87,9 +87,29 @@ const inlineStyles = {
   }
 };
 
+const sortByName = (a,b) => {
+  const aName = `${a.first_name} ${a.last_name}`.toLowerCase();
+  const bName = `${b.first_name} ${b.last_name}`.toLowerCase();
```
mienaikoe: remove this from EmailUtils (see previous commenT). You can do a one-liner if you add full name as a method on the mongoose model.

mienaikoe:
Also please remember to do merge conflicts

PR #81, opened 2019-04-20
Require all users to verify email

k3ntako:


k3ntako:
- User will not be able to access the website without verifying their email. 
  - They will be shown a page asking them to verify
- Updated verification email
- Updated verification checks (removed many on front end)

client/components/Nav/Nav.js
```diff
@@ -89,8 +89,7 @@ class Nav extends Component {
             <MenuItem
                 caption='Edit Account'
                 icon="edit"
-                onClick={this.linkToMe}
-                disabled={!me.is_verified}/>
```
mienaikoe: fwiw, I think you should keep verification checks in just in case we goof up real bad somehow.

client/contexts/MeContext.js
```diff
@@ -59,6 +61,8 @@ class MeProviderComponent extends React.Component {
       default:
         if( !this.state.me ){
           this.props.history.replace('/login');
+        }else if( !this.state.me.is_verified ){
+          this.props.history.replace('/unverified');
```
mienaikoe: This would prevent any access to logout. What if someone else logins on your device and then never verifies. You would be eternally locked out of the site.

client/models/Person.js
```diff
@@ -84,9 +84,10 @@ class Person extends Model {
   static me(){
     return Fetcher.get('/api/me').then( personJson => {
       let person = new Person(personJson);
-      if( person.team ){
+      if( person.team && person.team.members ){
         person.team.members = person.team.members.sort(sortByName);
```
mienaikoe: never noticed this before, but this sorting should be done on the backend.

client/contexts/MeContext.js
mienaikoe: nvm. I see you thought about that already.

client/models/Person.js
mienaikoe: perhaps a different PR unless it's easy to do here.

client/pages/Auth/Unverified/index.js
```diff
@@ -0,0 +1,59 @@
...
+  return <div className={styles.auth}>
+    <div className={styles.authOverlay}>
+      <div className={styles.authForm}>
+        <div className={styles.logo} />
+        <div className={styles.verifyMessage}>
+          <h2>Verify Your Email</h2>
+          <p>
+            <span>{"We've"} sent an email to <strong>{email}</strong> with a verification link. </span>
+          </p>
+          <p>
```
mienaikoe: These can be the same paragraph to save space and connect the two sentences visually.

client/pages/Auth/Unverified/index.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  const email = props.me.email || "you";
+
+  return <div className={styles.auth}>
+    <div className={styles.authOverlay}>
+      <div className={styles.authForm}>
+        <div className={styles.logo} />
+        <div className={styles.verifyMessage}>
+          <h2>Verify Your Email</h2>
```
mienaikoe: I think an h3 might make it a little less "demanding" sounding.

client/pages/Auth/index.css
```diff
@@ -102,6 +102,7 @@
 
 .verifyMessage > h2{
   text-align: center;
+  margin-top: 0;
 }
 
 .verifyMessage > p{
```
mienaikoe: I think the p in all auth forms could stand to have a bit more breathing room in the `line-height` department

server/emails/resend_verification/html.pug
```diff
@@ -5,10 +5,12 @@ body
 
   p
     div You have requested to resend the verification email.
-    If you believe someone else has done this,
-    please contact Jesse or Kentaro and disregard this message.
+    Please click on the following link, or copy the full URL in to the browser.
```
mienaikoe: Smart.

client/pages/Auth/Unverified/index.js
k3ntako: in addition to making it h3, what if i added a "Please" or "Let's" in front?

k3ntako: also, i feel like h3 is too small.

mienaikoe: Good with Let's.

How about something between h2 and h3. Just like a div that you style specifically for auth.

client/models/Person.js
k3ntako: I just copied the sort to the backend. I wanted to sort it in the populator, but I couldn't figure out how sort by full name in lowercase.

mienaikoe: https://stackoverflow.com/questions/52015712/mongodb-sort-with-case-insensitive-manner/52015956

k3ntako: I saw this earlier, but I couldn't really get it to work.

k3ntako: Decided to put this in another PR. I have a lot more related issues to resolve.

mienaikoe: ah yea. not easy to do in conjunction with a populate call. fine to boot to a diff PR

PR #82, opened 2019-04-20
Location form bug fix

k3ntako:
1. Submit button disabled when existing location selected
  - Unless user selected a new location on Googles Maps, Submit button would remain disabled.
2. Google Maps was receiving a boolean for onClick if `fungible` was false. Should be function.


client/components/LocationForm/index.js
```diff
@@ -64,7 +64,14 @@ export default class LocationForm extends React.Component {
     }
     const location = this.props.locations.find( location => location._id === locationId );
     if( location ){
-      this.setState( stateFromLocation(location) );
+      let newState = stateFromLocation(location);
+      const newValid = isValid(newState);
```
mienaikoe: `stateFromLocation` already evaluates for isValid. Remove this.

client/components/LocationForm/index.js
```diff
@@ -64,7 +64,14 @@ export default class LocationForm extends React.Component {
...
+
+      if( this.state.valid !== newValid ){
+        this.props.onValidChange(newValid)
```
mienaikoe: I would call this after setState is done.

client/components/LocationForm/index.js
```diff
@@ -64,7 +64,15 @@ export default class LocationForm extends React.Component {
     }
     const location = this.props.locations.find( location => location._id === locationId );
     if( location ){
-      this.setState( stateFromLocation(location) );
+      let newState = stateFromLocation(location);
+
+      if( this.state.valid !== newState.valid ){
+        return this.setState( newState,
+          () => this.props.onValidChange(newState.valid)
```
k3ntako: set callback in if statement

setState only once

PR #83, opened 2019-04-21
sorting changes, bug fixes and other

k3ntako:
Unused methods in Boat Controller.
Moved sort to back-end, if possible
Few other changes noted as comments.


server/controllers/SessionController.js
```diff
@@ -178,12 +179,23 @@ const SessionController = {
...
+      if( req.session.is_verified ){
+        populator = COACH_POPULATOR;
+      }
+
+      // refresh person to populate appropriate fields
+      // required for both coaches & members (e.g., add invites and remove auth)
+      person = await Person.
+        findOne({ _id: person._id }).
+        select(SAFE_SELECT).
+        populate(populator);
```
k3ntako: we need to repopulate so we can use `select` and use the appropriate populator

client/pages/Email/Recipients.js
```diff
@@ -98,6 +97,25 @@ class Recipients extends Component {
         }
       });
 
+      let members = [];
+      let invites = [];
+      if( props.isCoach ){
+        props.email.recipients.forEach(person => {
+          if(typeof person.first_name === 'string'){
```
mienaikoe: I'd rather you check if the person has a team set to determine if they're an invite or not.

This also would make nonMembers -> Person change more difficult.

mienaikoe: Also, technically, if you're sorting by name on the backend, all the undefined names (invites) would be sorted already.

client/pages/Email/Recipients.js
```diff
@@ -98,6 +97,25 @@ class Recipients extends Component {
...
+          }else{
+            invites.push(person);
+          }
+        })
+      }else{
+        members = recipients;
+      }
+
+      invites = invites.sort( (a,b) => a.email > b.email ? 1 : -1 );
+      members = members.sort( (a,b) => a.searchableName > b.searchableName ? 1 : -1 );
```
mienaikoe: I thought they were sorted on the backend already.

server/controllers/EmailController.js
```diff
@@ -66,24 +66,30 @@ const EmailController = {
...
-      const recipientIds = email.recipients;
       email = await email.populate(DEEP_POPULATOR);
 
       const recipients = email.recipients;
-      let invites = [];
       let members = [];
+      let invites = [];
+      if( req.session.is_coach ){
+        recipients.forEach(person => {
+          if(typeof person.first_name === 'string'){
```
mienaikoe: check for team, not first name.

server/controllers/EmailController.js
```diff
@@ -66,24 +66,30 @@ const EmailController = {
...
+            members.push(person);
+          }else{
+            invites.push(person);
+          }
+        })
+      }else{
+        members = recipients;
+      }
```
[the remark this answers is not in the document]
mienaikoe: There's no reason to do any of this if you're just going to do the same thing on the frontend.

server/controllers/EventController.js
```diff
@@ -73,21 +72,9 @@ const EventController = {
...
-      if( !req.session.is_verified ){
-        populator.forEach((path, idx) => {
-          if(path.select && path.model === 'Person'){
-            populator[idx].select += UNVERIFIED_SELECT;
-          }
-          if(path.populate && path.populate.select && path.populate.model === 'Person'){
-            populator[idx].populate.select += UNVERIFIED_SELECT;
-          }
-        })
-      }
```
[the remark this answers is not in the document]
mienaikoe: Keep the check. Any malicious user could potentially start making changes if they learn our API.

server/controllers/SessionController.js
```diff
@@ -26,6 +27,13 @@ const DEEP_POPULATOR = {
   ]
 };
 
+//removes reference
+//Object.assign is shallow, so populate requires '.slice()'
+let DEEP_POPULATOR = Object.assign({}, COACH_POPULATOR);
+DEEP_POPULATOR.populate = COACH_POPULATOR.populate.slice();
+DEEP_POPULATOR.select = "-invites";
+DEEP_POPULATOR.populate.pop();
```
mienaikoe: would prefer if you used `unshift` with index 2. In case we add another item to that list and we don't instinctively think to modify this line.

server/controllers/SessionController.js
```diff
@@ -107,11 +115,6 @@ const SessionController = {
               $pull: { invites: person._id },
               $push: { members: person._id }
             })
-            // refresh person because its populated team is now outdated
-            person = await Person.
```
mienaikoe: Why did you delete this? It clearly states that it's needed

client/pages/Email/Recipients.js
k3ntako: order gets messed up because user can add/remove recipients.

server/controllers/SessionController.js
```diff
@@ -121,14 +124,21 @@ const SessionController = {
...
       req.session.person_id = person._id
       req.session.is_verified = person.is_verified;
+      req.session.is_coach = false;
+
+      // remove any sensitive data
+      // if user had a team, the team is now outdated
+      person = await Person.
+        findOne({ _id: person._id }).
+        select(SAFE_SELECT).
+        populate(DEEP_POPULATOR);
```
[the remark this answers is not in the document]
mienaikoe: I would put what you said in this PR comment into a code comment. "the team is now outdated" isn't real specific.

server/controllers/SessionController.js
```diff
@@ -178,12 +179,23 @@ const SessionController = {
         }
       }
 
-      return person;
-    }).then( person => {
-      person.set('auth', null)
+      let populator = DEEP_POPULATOR;
+      if( req.session.is_verified ){
+        populator = COACH_POPULATOR;
```
mienaikoe: if you're going to keep using COACH_POPULATOR for things that are unrelated to being a coach, you ought to call it something else.

server/controllers/SessionController.js
mienaikoe: You can remove pieces of information without doing another database call.

server/models/NonMember.js
```diff
@@ -20,7 +20,9 @@ const NonMemberSchema = mongoose.Schema({
                       }
 });
 
-
+NonMemberSchema.methods.searchableName = function() {
+  return `${this.first_name.toLowerCase()} ${this.last_name.toLowerCase()}`;
```
mienaikoe: What if first_name or last_name are null?

server/models/Person.js
```diff
@@ -121,7 +121,7 @@ PersonSchema.pre('remove', function(next){
   }).then( next );
 });
 
-PersonSchema.methods.searchableName = function(cb) {
+PersonSchema.methods.searchableName = function() {
   return `${this.first_name.toLowerCase()} ${this.last_name.toLowerCase()}`;
```
mienaikoe: same here. what if first or last name are null (esp in the case of invites)

server/controllers/EmailController.js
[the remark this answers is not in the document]
k3ntako: this is where most of the sorting happens.

i only do it on the front end when the user sends the email. that's necessary because the order of people changes when they check/uncheck recipients. 

Alternatively, we could remove it from the front end and return a sorted email from the backend.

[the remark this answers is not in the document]
mienaikoe: I still don't get it. Let's sit down and talk about this sometime.

PR #84, opened 2019-04-21
Fixes

k3ntako:


server/controllers/SessionController.js
```diff
@@ -178,9 +179,11 @@ const SessionController = {
         }
       }
 
-      return person;
+      return Person.
+        findById(person._id).
+        select(SAFE_SELECT).
+        populate(DEEP_POPULATOR).then(person => person);
```
k3ntako: this is returning all info on a person except auth. 

can we use `select()` without a database call? My understanding was that it does not work.

server/controllers/EmailController.js
```diff
@@ -73,17 +73,25 @@ const EmailController = {
...
+        recipients.forEach(person => {
+          if(typeof person.first_name === 'string'){
+            members.push(person);
+          }else{
+            invites.push(person);
+          }
+        })
+      }else{
+        members = recipients;
+      }
```
k3ntako: This is a duplicate code, but it prevents us from having to send an email back to the front-end.

https://github.com/mighty-speed-krush/Mighty-Speed-Krush/pull/83/files/9bc1a4da016129237ab3c6af986c08a78e9061bc#r277154982

server/models/NonMember.js
```diff
@@ -20,7 +20,13 @@ const NonMemberSchema = mongoose.Schema({
                       }
 });
 
+NonMemberSchema.methods.searchableName = function() {
+  let { first_name, last_name } = this;
+  first_name = first_name ? first_name : "";
+  last_name = last_name ? last_name : "";
 
+  return `${first_name} ${last_name}`.trim().toLowerCase();
+};
```
k3ntako: currently, we are sorting using `nonMember.searchableName()`, however the instance method doesn't exist.

client/pages/Email/Recipients.js
```diff
@@ -91,12 +91,32 @@ class Recipients extends Component {
...
+          }
+        })
+      }else{
+        members = props.email.recipients;
+      }
+
+      invites = invites.sort( (a,b) => a.email > b.email ? 1 : -1 );
+      members = members.sort( (a,b) => a.searchableName > b.searchableName ? 1 : -1 );
+
+      const includedMembers = invites.concat(members);
```
k3ntako: without the sorting here and in the EmailController, the recipients table will not be alphabetized.

server/controllers/EmailController.js
```diff
@@ -68,22 +68,32 @@ const EmailController = {
 
       const recipientIds = email.recipients;
       email = await email.populate(DEEP_POPULATOR);
+      const team = await Team.findById(email.team);
```
k3ntako: we could populate them team on them email, because every email has a team. But every email we return will have a team populated. I though that this would be faster, especially if the user has bad connection.

Or we could conditionally populate. But, i thought that would be harder to read than this.

client/pages/Email/Recipients.js
mienaikoe: I don't understand why.

k3ntako: First time it sorts when it fetches.
email.recipients is sorting as “invites” then “members”

If I include another invite it gets pushed to the end of email.recipients. Now the order is invites => members => invite. 

I maintain this original order while the person edits the email. Mostly so if the user accidentally checked/unchecked someone, they don’t have to go looking for them. 

What this is doing is sorting the recipients which is now not sorted when someone clicks send.

I guess one issue is that if the email was sent prior to loading the page, this shouldn’t be triggered

PR #85, opened 2019-04-21
Enter key to proceed to next input

k3ntako:
On mobile hitting enter generally brings the user to the next field. 
I think this is important on a mobile device where tapping the next field is difficult/error prone.

Hitting enter while on the last input will submit form.


client/pages/Auth/Join/JoinPage.js
```diff
@@ -93,17 +101,25 @@ class JoinPage extends React.Component {
...
             value={this.state.firstName}
-            onChange={firstName=>this.setState({firstName})}/>
+            onChange={firstName=>this.setState({firstName})}
+            onKeyPress={(ev) => this.listenForEnter(ev, null, this._input2)}/>
           <Input type='text' label="Last Name"
+            ref={ el => { this._input2 = el } }
             value={this.state.lastName}
-            onChange={lastName=>this.setState({lastName})}/>
+            onChange={lastName=>this.setState({lastName})}
+            onKeyPress={(ev) => this.listenForEnter(ev, this._input1, this._input3)}/>
```
mienaikoe: You can make this a wrapper component for `Input` instead of repeating `onKeyPress=...` a bunch of times.

PR #86, opened 2019-04-22
Removed prompt after deleting email

k3ntako:
It's not big, but I thought you might have some opinions about it.


client/pages/Email/index.js
```diff
@@ -78,8 +79,14 @@ class TeamEmail extends Component {
     this.setState({ isSaved: true, email, isSaving: false });
   }
 
-  setVerb(field, isSending, callback){
-    this.setState({ [field]: isSending }, callback);
+  setVerb(field, newVerbState, callback = resolve => resolve()){
```
mienaikoe: resolve isn't defined here. If you don't pass a callback, you would be calling undefined.

k3ntako: if callback isn't defined, `resolve => resolve()` is called instead?

`resolve` is passed in when the callback is called? Maybe it'd be clearer if I called the argument something else? like this: `setVerb(field, newVerbState, callback = r => r()){`

mienaikoe: why do you need to pass in a callback in the first place? Why can't the caller resolve the promise?

k3ntako: the `setState` returns `undefined` regardless of what the callback returns.
https://github.com/facebook/react/issues/2642
https://stackoverflow.com/questions/54273673/react-setstate-callback-return-value

I did try to make `setVerb` an `async` function but then I ran into another issue (`regeneratorRuntime is not defined`). Seems like we can't make instance methods into async functions, unless we add a babel plugin. 
https://stackoverflow.com/questions/49253746/error-regeneratorruntime-is-not-defined-with-babel-7

mienaikoe: Keep the promise. I mean anything that calls setVerb should handle the promise instead of supply a callback

client/pages/Email/EditEmail.js
```diff
@@ -68,20 +69,23 @@ class EditEmail extends Component {
...
   deleteEmail(){
-    return this.props.email.del().then(() => {
-      this.props.history.push(`/#email`);
+    return this.props.setIsDeleting(true).then(() => {
+      return this.doDeleteEmail();
     });
   }
 
+  doDeleteEmail(){
+    return this.props.email.del();
```
mienaikoe: don't you need to go back to the email list?

client/pages/Email/index.js
```diff
@@ -54,6 +55,10 @@ class TeamEmail extends Component {
     }
   }
 
+  componentWillUnmount(){
+    window.onbeforeunload = undefined
```
mienaikoe: Is there a way to set this with `addEventListener`?

client/pages/Email/EditEmail.js
k3ntako: The `<DeleteConfirmation>` component takes care of that.

PR #87, opened 2019-04-22
Minor Email Frontend refactor

mienaikoe:
- Putting subject into the heading title
- Putting the sent time as a note box on the top of the email
- Dividing View and Edit Logic on email form


client/pages/Email/ViewEmail.js
```diff
@@ -0,0 +1,26 @@
+import React, {Component} from 'react';
+import Card from 'react-toolbox/lib/card';
+import DraftHelpers from './../../components/Draft-js/DraftHelpers';
+import Subheading from '../../components/Subheading';
```
k3ntako: Not using `DraftHelpers` and `Subheading`

k3ntako: Not using `DraftHelpers` and `Subheading`

client/pages/Email/index.js
```diff
@@ -154,7 +162,8 @@ class TeamEmail extends Component {
       <Prompt
         when={ email && !email.sendTime && !isSaved || isSaving || isSending }
         message='You have unsaved changes, are you sure you want to leave?' />
-      <Heading title="Team Email" crumbs={crumbs} />
+      <Heading title={this.state.newSubject || (email && email.subject)}
```
k3ntako: don't think we should have a placeholder?

client/pages/Email/index.js
```diff
@@ -118,18 +136,8 @@ class TeamEmail extends Component {
...
-        newSubject={newSubject}
-        emailChangeHandler={this.emailChangeHandler.bind(this)}
-        subjectChangeHandler={this.subjectChangeHandler.bind(this)}
-        isSending={isSending}
-        setIsSending={this.setVerb.bind(this, "isSending")}
-        isSaved={isSaved}
-        setIsSaved={this.setIsSaved.bind(this)}
-        isSaving={isSaving}
-        setIsSaving={this.setVerb.bind(this, "isSaving")} />
+      <Subheading label="Message" />
```
k3ntako: Not sure how I feel about this. A part of me feels like this is not necessary, but it is better for consistency?

mienaikoe: yea it could just be an h3. But I was thinking like sometime down the line if we want to restyle all subheadings.

Also for readability, seeing Subheading a bunch of times I think is easier to read than h3 and then subheading in the same file.

k3ntako: oh I meant the heading. It's clear that it's the email editor?

mienaikoe: Everything is much clearer to the author.

Whenever I go to the email form, I'm always kind of looking for something to look at first and mentally associate with, but there's nothing there to anchor on.

client/pages/Email/index.js
k3ntako: sorry don't think I was clear enough. I mean if there is no subject, should we have a place holder like: "Team Email". The size of the Header changes dramatically when the user starts typing

mienaikoe: Maybe "New Email" ?

k3ntako: yeah that sounds better!

client/pages/Email/index.js
k3ntako: okay that makes sense.

PR #90, opened 2019-04-24
Confirmation for navigating away from edit event/series

k3ntako:


client/pages/EventForm/EventForm.js
```diff
@@ -97,11 +122,13 @@ class EventForm extends Component {
 
     if(this.props.match.path.includes("/edit")){
       this.state.event.update(name, datetime, location, draftNote, assignments).then( (event) => {
+        window.removeEventListener('beforeunload', this.beforeunload);
```
mienaikoe: you can just let componentWillUnmount handle this

client/pages/EventForm/EventForm.js
```diff
@@ -97,11 +122,13 @@ class EventForm extends Component {
...
         this.props.history.push(`/events/${this.props.match.params.id}`);
       });
     }else {
       const assignmentNames = assignments.map( assignment => assignment.name );
       Event.create(name, datetime, location, draftNote, assignmentNames).then( (event) => {
+        window.removeEventListener('beforeunload', this.beforeunload);
```
mienaikoe: same as above

client/pages/SeriesForm/SeriesForm.js
```diff
@@ -181,6 +209,7 @@ class SeriesForm extends Component {
                 key="draft"
                 label="Notes for Event"
                 onChange={this.eventNoteChangeHandler}
+                setHasChanged={this.setHasChanged.bind(this)}
```
mienaikoe: do the binding in the constructor so you don't have to call bind a bunch of times at render

client/components/EventTextEditor/index.js
```diff
@@ -4,8 +4,16 @@ import styles from './EventEditor.css';
 
 
 export default (props) => {
+  const onChange = (editorState) => {
+    if(editorState.getSelection().getHasFocus()){
+      props.setHasChanged();
+      return props.onChange(editorState)
```
mienaikoe: can't props.onChange call setHasChanged on its own? I don't think you need to call them both back-to-back

mienaikoe:
Also I suggest making the target of this PR dev instead of 1.1.X

We're not demoing this today.

client/components/EventTextEditor/index.js
k3ntako: I moved `setHasChanged()` up one level. is that what you meant?

mienaikoe: depends on what you mean by level

mienaikoe: ah yep. that's what I meant

client/pages/EventForm/EventForm.js
```diff
@@ -84,6 +84,7 @@ class EventForm extends Component {
   }
 
   eventNoteChangeHandler(note){
+    this.setHasChanged();
     this.setState({ note });
```
mienaikoe: Doesn't this effectively do two `setState` calls? Collapse this to one. It looks like you did this for all the other handlers.

k3ntako: "did this" meaning I did collapse it, right?  Meaning this is the only place to fix it?

PR #94, opened 2019-04-25
Member Remove

mienaikoe:
- Moves invite/uninvite to TeamController
- Adds an endpoint for removing a member
- Adds an action to roster to remove a member


server/controllers/TeamController.js
```diff
@@ -62,6 +78,205 @@ const TeamController = {
     next();
   },
 
+  async invitePerson(req, res, next){
```
k3ntako: I think we should call this `invitePeople`

server/controllers/TeamController.js
```diff
@@ -62,6 +78,205 @@ const TeamController = {
...
+          },{
+            non_attendees: person_id,
+          }
+        ]
+      });
+      
+      // Pull them from all boats they appear on
+      await Boat.updateMany({
+        event: {
+          $in: attendingEvents//.map( event => event._id )
```
k3ntako: you have a comment here.

server/routes.js
```diff
@@ -60,17 +60,18 @@ router.route( '/me/settings').
...
-  delete(                                requireCoach,    PersonController.uninviteFromMyTeam);
 router.route( '/people/:person_id').
   get(                                   requireVerified, PersonController.find).
   put(                                   requireCoach,    PersonController.update);
 
 router.use(   '/team',                   requireVerified);
 router.route( '/team/mine').
   get(                                                    TeamController.mine);
+router.route( '/team/invite').
+  post(                                  requireCoach,    TeamController.invitePerson).
```
k3ntako: same as above, `invitePeople` instead of `invitePerson`

server/tests/SessionControllerTest.js
```diff
@@ -106,7 +105,7 @@ describe('SessionController', function() {
         }
       }
 
-      PersonController.inviteToMyTeam({
+      TeamController.invitePerson({
```
k3ntako: same as above, `invitePeople` instead of `invitePerson`

server/tests/TeamControllerTest.js
```diff
@@ -36,4 +37,106 @@ describe('TeamController', function() {
     })
   })
 
+
+  let invited_person_id = null;
+
+  describe('#invitePerson', function(){
```
k3ntako: same as above, `invitePeople` instead of `invitePerson`

server/tests/TeamControllerTest.js
```diff
@@ -36,4 +37,106 @@ describe('TeamController', function() {
...
+        }
+      },{
+        json: function(response){
+          assert(response.okay);
+          Team.findById(coach_session.team_id).then( team => {
+            const foundMember = team.members.find( mpid => mpid.equals(newPerson._id) )
+            assert(!foundMember)
+            done();
+          }, done)
+        }
```
k3ntako: maybe this is too much, but ideally we would test if they were removed from events/boats/assignments.

server/tests/TeamControllerTest.js
```diff
@@ -36,4 +37,106 @@ describe('TeamController', function() {
     })
   })
 
+
+  let invited_person_id = null;
+
+  describe('#invitePerson', function(){
+
+    it('should invite a person to my team', function(done){
+      TeamController.invitePerson({
```
k3ntako: same as above, `invitePeople` instead of `invitePerson`

mienaikoe:
Oh damn that's a great Idea to move it to the person page.

server/tests/TeamControllerTest.js
mienaikoe: good call on this. There were a few things faulty about the test.

mienaikoe:
ready for 2nd review

client/pages/Person/index.css
```diff
@@ -5,19 +5,14 @@
...
-.coach{
-  width: 320px;
 }
-
-.card{
-  padding: 2rem;
+.sidebar{
+  width: 240px;
+  margin-left: 2rem;
 }
```
k3ntako: You're not using this. I think you forgot to add the classname. 

Is this why the cards are super wide?

mienaikoe: Yea so I decided to move the coach sections down and get rid of the sidebar, caus there really isn't any tertiary information that's important to be above the fold on this page.

Also coach just asked if it's possible to "archive" (basically putting people under the line) so I'll see if I can bake that into this sprint.

mienaikoe: ^ to clarify, I'm not going to put that extra feature into this PR

PR #95, opened 2019-04-25
Map Refactor

mienaikoe:
- Fixing Map Clicking Issues
- Modifying Map Cursor
- Fixing Map InfoWindow recalling
- Fixing Event Submit navigation block


client/components/EventMap/index.js
```diff
@@ -63,47 +63,35 @@ class EventMap extends Component {
...
-      const center = { lat, lng };
-      let markerCenter = center;
-
-      if(!props.location.latitude){
-        markerCenter = null
+  static getDerivedStateFromProps(props, state={}){
+    if( props.location ){
+      if( !state.location || state.location !== props.location ){
+        let lat = props.location.latitude || state.center.lat;
+        let lng = props.location.longitude || state.center.lng;
```
k3ntako: someone might try 0 for coordinates. wouldn't that cause some issues?

k3ntako: someone might try 0 for coordinates. wouldn't that cause some issues?

client/components/EventMap/index.js
```diff
@@ -133,11 +116,12 @@ class EventMap extends Component {
...
-      center: center,
-      markerCenter: center,
-      zoom: 16
-    });
+    // this.setState({
+    //   center: center,
+    //   markerCenter: center,
+    //   zoom: 16,
+    //   infoWindowOpen: true,
+    // });
```
k3ntako: remove comment

client/components/EventMap/index.js
```diff
@@ -63,47 +63,35 @@ class EventMap extends Component {
...
-      if(!props.location.latitude){
-        markerCenter = null
+  static getDerivedStateFromProps(props, state={}){
+    if( props.location ){
+      if( !state.location || state.location !== props.location ){
+        let lat = props.location.latitude || state.center.lat;
+        let lng = props.location.longitude || state.center.lng;
+
+        const center = { lat, lng };
+        let markerCenter = center;
```
k3ntako: you should be okay with `const`

client/components/EventMap/index.js
```diff
@@ -194,25 +204,15 @@ class EventMap extends Component {
...
-    if(markerCenter && location && typeof(location.latitude) !== 'undefined' ){
-      marker = <Marker
-          position={markerCenter}
-          draggable={fungible}
-          onDragEnd={this.onMarkerDragged}
-        >
-        <InfoWindow>
+    if(markerCenter ){
+      let infoWindow;
+      if( this.state.infoWindowOpen && location && typeof(location.address) !== 'undefined' ){
```
k3ntako: is there are a benefit for checking for `typeof(...) !== undefined` instead of checking if it's not falsy? 

We could do `&& location.address` instead of `typeof(location.address) !== 'undefined'`. I'd think that checking if it's falsy would allow us to catch more unwanted values. 

Is there a time when null (or something else falsy) should be accepted?

client/components/EventMap/index.js
mienaikoe: Lol yea I was thinking it would be such a slim chance that someone would want to go to [Null Island](https://en.wikipedia.org/wiki/Null_Island) but I guess I shouldn't be surprised if they do.

client/components/EventMap/index.js
mienaikoe: I think I was just worried that the address would sometimes be null intentionally like if someone decides to paddle in the middle of nowhere? But yea I have no clue what google would do about that tbh. I think it would still just try to geocode to the nearest address, but the lat/lon would still be where you click. I'll make it simpler as you suggested.

mienaikoe:
ready for re-review

PR #96, opened 2019-04-25
Attendance Override

mienaikoe:
- Adds No Response to atttendance tab on event page
- Adds checkboxes and buttons for overriding attendance when you are the coach

TODO: Make checkbox tables more generic, and have them use the scrollable table component.


client/models/Event.js
```diff
@@ -47,6 +47,25 @@ class Event extends Model {
     })
   }
 
+  overrideAttendance( people, is_positive ){
```
k3ntako: wouldn't `is_attending` be more clear than `is_positive`?

client/pages/Event/EventAttendance.js
```diff
@@ -1,94 +1,263 @@
...
+    const { team } = me;
+    const isCoach = me.isCoach();
+
+    const respondedIds = event.attendees.
+      concat(event.non_attendees).
+      map( person => person._id)
+
+    let noResponses = team.members.filter( member => {
+      return respondedIds.indexOf(member._id) === -1 &&
+        team.coaches.indexOf(member._id) !== -1;
```
k3ntako: `team.coaches` is an array of `Person`s. This has to be converted to an array of `_id`s.

client/models/Event.js
mienaikoe: yes, but I decided to use the same boolean as the `attend()` function so as to avoid confusion, since they effectively do the same things. Maybe I'll convert them both to is_attending

client/pages/Event/EventAttendance.js
mienaikoe: whoa good catch!

mienaikoe:
ready for re-review

client/pages/Event/EventAttendance.js
```diff
@@ -1,94 +1,176 @@
...
+          Mark As Attending
+        </Button> }
+      </Subheading>
+      { this.renderNonAttendees(event.non_attendees, "checkedNonAttendees") }
+      <Subheading label={`No Response (${noResponses.length})`}>
+        { isCoach && <Button icon="check"
+            primary
+            onClick={this.overrideAttendance.bind(this, "checkedNoResponses", true)}
+            disabled={!this.state.checkedNoResponses.length}>
+          Mark As Attending
```
k3ntako: This might be something we should ask coach, but do you think coach wants to change non-respondents to "Attending" or "Not Attending"? 

Maybe he's just gonna assume that no response means not attending. But, if he wants everyone to be either attending or not attending, then most non-respondents would probably be not attending.

I approved the PR because I think there's a strong argument to be made for keeping it this way. Just wanted to ask you the question.

mienaikoe: Well, I don't know for sure what he would say, but I think he would prefer to know who hasn't responded just in case he needs to send a strongly worded email lol.

mienaikoe: He also specifically asked for management of attendance, because oftentimes there are discrepancies between who actually attended and who marked it on the spreadsheet.

mienaikoe: Oh you're asking which "direction" they should default to? hmm. I think this would be the direction, because it's not often you want people in the no-response group to be moved to the did-not-attend group because they mean almost the same thing. No-Response just has the added layer of being socially distant from the team.

mienaikoe: And either way, coach can always move from attending to not-attending, so that movement would just take two clicks.

PR #98, opened 2019-04-29
Moving Coach Fields to a nested Coach Fields object

mienaikoe:
- Removing preferred side and position as we aren't reflecting those fields anywhere atm, and I think coach doesn't want the team to think about those. If we need them again later, they're easy to add to coach_fields.

- Moving sex, weight, and erg times into a new coach fields object on the user. This should simplify selects and clarify the purpose of those fields

- Did not remove the existing fields for coach fields at the top level, as they are needed for migration. Once you approve of the data model of this PR, I will migrate the prod database (essentially data copy) and then add a new commit to this PR to remove the existing fields.


server/models/Person.js
```diff
@@ -53,18 +41,25 @@ const PersonSchema = mongoose.Schema({
...
   birthday:             { type: Date, default: null },
   cpr_certified:        { type: Boolean, default: false, required: true },
   shirt_size:           { type: String, default: null, enum: SHIRT_SIZES },
+  coach_fields: {
+    weight:               { type: Number, default: null, min: 0 }, // grams
+    sex:                  { type: String, default: null, enum: SEXES },
+    erg200m:               { type: Number, default: null, min: 0 }, // seconds
+    erg500m:               { type: Number, default: null, min: 0 }, // seconds
+    erg2000m:              { type: Number, default: null, min: 0 }, // seconds
+  },
```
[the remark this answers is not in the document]
mienaikoe: Yea I was thinking about that in https://trello.com/c/9CMcMRYk/288-records-one-pager

I think the fitness fields will inevitably turn into an arbitrary-key map based on the fitness fields of the team. Just learned about these: https://mongoosejs.com/docs/schematypes.html#maps

Maybe I should make fitness_fields a map then?

server/migrations/6_RefactorCoachFields.js
```diff
@@ -0,0 +1,27 @@
...
+  const peoples = await Promise.all(allPeople.map( person => {
+    if( !person.coachFields ){
+      person.coachFields = {
+        weight: person.weight,
+        sex: person.sex,
+        erg200m: person.erg_times['200m'],
+        erg500m: person.erg_times['500m'],
+        erg2000m: person.erg_times['2000m'],
+      }
+      return person.save()
```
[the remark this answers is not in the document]
mienaikoe: ah shoot. I changed the case in the migration. I thought it did work, but maybe I was using a different case before...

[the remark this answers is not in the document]
mienaikoe: I don't need an await here because I'm awaiting the Promise.all

server/models/Person.js
```diff
@@ -53,18 +41,25 @@ const PersonSchema = mongoose.Schema({
                         },
   phone:                { type: String, default: null },
   location:             { type: String, default: null },
-  weight:               { type: Number, default: null, min: 0 }, // min doesn't work
-  preferred_position:   { type: String, default: null, enum: PREFERRED_POSITIONS },
-  preferred_side:       { type: String, default: null, enum: PREFERRED_SIDES },
   birthday:             { type: Date, default: null },
   cpr_certified:        { type: Boolean, default: false, required: true },
   shirt_size:           { type: String, default: null, enum: SHIRT_SIZES },
```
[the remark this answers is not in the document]
mienaikoe: Hmm. Yea I always assumed it was in case someone besides the coach was buying the jerseys. but... That's a rare case, and I absolutely understand that because it's almost like weight. I'll move it into coach_fields.

server/models/Person.js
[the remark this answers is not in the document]
mienaikoe: Called it latest_metrics because I want it to reflect how we might approach records/metrics in the future.

server/models/Person.js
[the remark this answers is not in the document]
mienaikoe: Okay, so since shirt_size is already on the top level, I'm considering just keeping it there, and adding it as an unselect later. coach_fields was meant to be fields on a person that only a coach could see or set. shirt_size is now a field that both coach and that person would see and the shirt_size can only be set by the person, not the coach.

Weight is a different story. On one hand, maybe it should be a metric to track it over time. On the other hand, weight is important for boat balance. On one hand, the coach needs to set it for it to be accurate. On the other hand, the person should be able to see their own weight.

For now I will keep weight in latest_metrics because only coach can set it, but coach and person should be able to see it.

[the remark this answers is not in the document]
mienaikoe: since coach_fields only has sex, and I doubt it will have many more fields, I'm going to move sex back to the top level.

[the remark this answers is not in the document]
mienaikoe: Also, coaches might consider sex to be user-settable in the future. So we'll see what happens.

server/migrations/6_RefactorCoachFields.js
```diff
@@ -0,0 +1,49 @@
...
+        weight: person.weight,
+        sex: person.sex,
+        shirt_size: person.shirt_size,
+      }
+      person.metrics = {
+        "weight": {
+          timestamp,
+          resistance: person.weight,
+          endurance: null
+        },
```
k3ntako: why do you put weight in metrics? we have it in `coach_fields`. Similar comment above.

server/migrations/6_RefactorCoachFields.js
```diff
@@ -0,0 +1,49 @@
...
+
+  const timestamp = new Date();
+
+  const peoples = await Promise.all(allPeople.map( person => {
+    if( !person.coach_fields || !person.metrics ){
+      person.coach_fields = {
+        weight: person.weight,
+        sex: person.sex,
+        shirt_size: person.shirt_size,
+      }
```
k3ntako: For some reason `person.coach_fields` isn't created after running the migration. The fake user does have a `sex`, `shirt_size`, and `weight` (top arrow).

However, the `metrics` is created (bottom arrow).

Screen Shot 2019-05-02 at 10 13 43 AM

client/pages/Person/SeatingForm.js
```diff
@@ -17,39 +17,47 @@ const SEXES = [
...
-      weight: UnitUtils.lbToGrams(this.state.weight),
-      preferred_side: this.state.preferred_side,
+    const params = {
       sex: this.state.sex,
-      erg_times: this.state.erg_times
-    }).then( () => {
+      metrics: {
+        weight: {
+          resistance: UnitUtils.lbToGrams(this.state.weight),
+        },
```
k3ntako: Shouldn't weight be outside of `metrics`?

server/models/Person.js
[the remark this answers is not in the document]
k3ntako: I think we should keep the weight separate from the fitness. He mentioned that he wants the fitness results to be visible by the team. And I think it doesn't make sense to have an `endurance` and `resistance` fields. 

I understand the benefits of having timestamps, so I think we could keep the weight object on the top level or in `coach_fields`.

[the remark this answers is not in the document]
mienaikoe: had a conversation about this.

- Make sex field `me`-settable, and visible only to coach-and-me
- Keep metrics, but take out endurance/resistance and just have one value

server/migrations/6_RefactorCoachFields.js
mienaikoe: had conversation

server/migrations/6_RefactorCoachFields.js
mienaikoe: coach_fields was removed in favor of weight in metrics, and sex at the top level

client/pages/Person/SeatingForm.js
mienaikoe: had conversation

client/pages/Person/PersonPage.js
```diff
@@ -14,6 +14,12 @@ import SeatingForm from './SeatingForm';
 import styles from './index.css';
 
 
+const SEX_LABELS = {
+  M: "Male",
+  F: "Female",
+  O: "Unspecified"
+};
```
k3ntako: Let's make this alphabetical order and put Female first.

server/migrations/6_RefactorCoachFields.js
```diff
@@ -0,0 +1,28 @@
...
+        "erg2000m": (person.erg_times['2000m'] || 0) * 1000,
+      }
+      return person.save()
+    } else {
+      return Promise.resolve()
+    }
+  }));
+
+  console.log(peoples);
+}
```
k3ntako: the migration keeps hanging for me and I can't figure out why. But when I run it with the tests, nothing fails.

`(node:34642) UnhandledPromiseRejectionWarning: ValidationError: Person validation failed: metrics.$*: Cast to number failed for value "5ccaf9d60531c87a9510edfa" at path "metrics.$*"`

It's possible that the migration from earlier messed up my data, but I can't see how that would affect it.

client/pages/Person/PersonPage.js
mienaikoe: This is kind of bikeshedding, but ok

server/migrations/6_RefactorCoachFields.js
mienaikoe: Ohh yea that could be why. To load a document, it has to fit the schema first. Maybe I should blow away all the metrics before running the update.

mienaikoe: actually blowing away metrics would destroy existing data for later migrations.

uhh... can you try blowing away metrics from compass?

k3ntako: It worked just fine!

PR #99, opened 2019-04-29
Populate and Select Refactor

mienaikoe:
- Moved Person selects into static strings on the Person Schema
- Specified these selects in populators

NOTE: Populate is largely an endpoint-dependent object, so it was not easy to refactor. However, selects are mostly viewer-dependent (barring some session-related things in join).


server/controllers/EmailController.js
```diff
@@ -8,7 +8,7 @@ const NonMember = require('../models/NonMember')
 const Email = require('../models/Email')
 const EmailUtils  = require('../utilities/EmailUtils');
 
-const SAFE_SELECT = "-auth -verification -weight -preferred_side -preferred_position -birthday -erg_times -sex";
+const SAFE_SELECT = "-weight -preferred_side -preferred_position -birthday -erg_times -sex";
```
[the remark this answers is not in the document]
mienaikoe: Actually I'm going to use Person.MEMBER_SELECT here and just remove SAFE_SELECT

I'll add shirt_size to MEMBER_SELECT

server/models/Person.js
```diff
@@ -81,6 +81,16 @@ const PersonSchema = mongoose.Schema({
...
+}, {
+  toJSON: {
+    transform: function(doc, ret){
+      delete ret.verification;
+      delete ret.auth;
+      if(ret.birthday){
+        ret.birthday.setFullYear(0);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
mienaikoe: toJSON is a function that gets called behind the scenes when JSON.stringify is applied to a mongoose object.

Yea this is overwriting the year... crap I need to find an alternative

server/models/Event.js
```diff
@@ -28,7 +28,16 @@ var EventSchema = new Schema({
     { type: ObjectId, ref: 'Assignment' }
   ],
   is_cancelled: { type: Boolean, required: true, default: false },
-}, { minimize: false }); //preserves empty objects
+}, { 
+  minimize: false,
+  toJSON: {
+    transform: (doc, ret, options) => {
+      if( doc.boats && !options.isCoach ){
+        ret.boats = ret.boats.filter( boat => boat.isVisible );
```
k3ntako: This is being used for the list of events. I remember talking about how we don't need to send over all this information for emails. Maybe that's also the case here. There's no reason to send over the whole list of events with boats, attendees and etc. 

Not sure if this is bad practice, but could we remove this `toJSON` for now and remove all the unnecessary fields within this sprint? 

If we aren't going to remove this, here are some issues:
I think you meant `doc.boats.filter()`, not `ret`. 
Also, `boat` is just an id, so `boat.isVisible` is going to be `undefined`.

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
...
+      "type": "node",
+      "request": "launch",
+      "name": "Launch Program",
+      "program": "${workspaceFolder}/server/index.js",
+      "env": {
+        "NODE_ENV": "development"
+      }
+    }
+  ]
+}
```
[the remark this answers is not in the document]
mienaikoe: ahh. yea.... ok.

server/models/Event.js
mienaikoe: I don't quite understand what you're requesting? Which fields need to be removed and how does that relate at all to this method?

Event's toJSON is used anytime an Event needs to be serialized. That includes when it's being listed and boats is just an array of ids, or when boats has been populated with the model. You're right that I need to check for isVisible, and for the existence of boats in general, but this toJSON is made specifically to handle the case of filtering out boats.

I do mean to reference ret, and not doc. Ret is an object resulting from pre-transforming the mongoose document. This method is for further transformation. Read more about it here: https://mongoosejs.com/docs/guide.html#toJSON

k3ntako: I thought this was just being used for the list of events. For the list, we don't need all the fields of an event. For example, the teampage does not show the user the boats. The same is true for the page with list of emails, and we talked about only responding with the `subject` and `sendTime`. I thought we could do something similar here.

Sorry, I assumed you wanted `doc` because you check for `doc.boats` in the `if` statement on line 36. Don't you wanna make sure `ret.boats` exists, not `doc.boats`, since that is what you're filtering through?

k3ntako: I think we can remove all that from the list of events in a different PR.

PR #103, opened 2019-05-06
Verify Email Test

k3ntako:


server/tests/SessionControllerTest.js
```diff
@@ -261,4 +358,84 @@ describe('SessionController', function() {
...
+          verification_token: verification_token + "a"
+        },
+      }
+      const res = {
+        json: function(response){
+          expect(response.okay).to.equal(false);
+          expect(response.error).to.equal('Provided verification token does not match');
+
+          Person.findOne({ email: email }).exec().then( person => {
+            expect(person.is_verified).to.equal(false);
```
mienaikoe: check the session as well

server/tests/SessionControllerTest.js
```diff
@@ -261,4 +358,84 @@ describe('SessionController', function() {
...
+          email: email,
+          verification_token: verification_token
+        },
+      }
+      const res = {
+        json: function(response){
+          assert(response.okay)
+
+          Person.findOne({ email: email }).exec().then( person => {
+            expect(person.is_verified).to.equal(true);
```
mienaikoe: check the session. especially here caus I think this is where the issues arose.

server/controllers/SessionController.js
```diff
@@ -349,7 +349,9 @@ const SessionController = {
         person = await person.save();
       }
 
-      req.session.is_verified = person.is_verified;
+      if( req.session.person_id ){
+        req.session.is_verified = person.is_verified;
+      }
```
k3ntako: We were setting `is_verified` for users who were not logged in. Thought we should make sure they're logged in.

server/tests/SessionControllerTest.js
```diff
@@ -424,10 +435,13 @@ describe('SessionController', function() {
...
 
           Person.findOne({ email: email }).exec().then( person => {
-            expect(person.is_verified).to.equal(true);
+            expect(person.is_verified).to.be.true;
             expect(person.verification.verification_hash).to.be.null;
             expect(person.verification.salt).to.be.null;
             expect(person.verification.token_expiration).to.be.null;
+            expect(req.session.is_verified).to.be.true;
+            expect(req.session.person_id).to.equal(unverified_session.person_id);
+            expect(req.session.team).to.equal(unverified_session.team);
```
mienaikoe: these two lines aren't testing anything. req.session is literally the same object as unverified_session. You should just test if they exist, because null can .equal null in chai.

server/tests/SessionControllerTest.js
```diff
@@ -261,4 +358,84 @@ describe('SessionController', function() {
...
+      verification_hash = passhashify(verification_token, salt);
+      const tomorrow = new Date().getTime() + (1000 * 3600 * 48);
+
+      const verification = {
+        verification_hash: verification_hash,
+        salt: salt,
+        token_expiration: tomorrow
+      }
+
+      TestHelper.createUnverifiedPerson(email, verification).then( session => {
```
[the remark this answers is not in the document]
mienaikoe: eh, nah. if you need to check the hash, keep it here.

server/controllers/SessionController.js
mienaikoe: hmm. Well should we trust the verification of an email if we can't make sure the person who is verifying is authenticated? Perhaps this endpoint should login the person.

mienaikoe:
huh.. TIL.

PR #104, opened 2019-05-07
Drag & Drop Seating!

mienaikoe:
- Adds a drag & drop seating solution for desktop
- Replaces the autocomplete modal with a list of attendees for mobile
- Adds sorting for team and event populators


client/components/BoatPanel/SeatingDock.js
```diff
@@ -0,0 +1,57 @@
...
+            0;
+      }).
+      map( (attendee, ix) => {
+        return <div key={attendee._id}
+            className={styles.dockSeat} 
+            onClick={this.onClick.bind(this, attendee)}>
+          <PaddlerClass 
+            {...this.props} 
+            person={attendee}
+            withBoats={true} />
```
k3ntako: what does `withBoats` mean?

client/pages/Boat/index.js
```diff
@@ -127,25 +80,30 @@ class BoatPage extends Component {
...
   }
 
   copyBoat(){
-    const { event } = this.state;
-    const templateBoat = event.boats[this.state.index];
-    return templateBoat.clone( event ).then( newBoat => {
+    const { event, boat } = this.state;
+    return boat.clone( event ).then( newBoat => {
       this.props.issueSnackbar("Boat Copied");
-      this.handleTabChange(event.boats.indexOf(newBoat));
```
k3ntako: could you remind me why we removed tabs? I remember talking about it as a way to save space.

client/components/BoatPanel/SeatingDock.js
```diff
@@ -0,0 +1,57 @@
...
+        </div>
+      })
+    
+    if( shownAttendees.length === 0 ){
+      shownAttendees = <div className={styles.seatingEmpty}>
+        All Attendees Seated
+      </div>
+    }
+
+    return <div className={styles.seatingDock}>
```
k3ntako: I think `.seatingDock` needs a `max-height`. I'm thinking it could be the same height as the boat, if possible.

client/components/BoatPanel/SeatingDock.js
mienaikoe: The boats that the person is already part of underneath their name. We had it in the autocomplete, but this pr removes the autocomplete, so I added it to Paddler. There are certain cases where it's useful, like the seating dock, and other places where it's not, like the boat panel.

client/pages/Boat/index.js
mienaikoe: Yea. I realized halfway through that it's annoying to have to click to see the balance and that in mobile, the tabs were unecessary because the seating dock wasn't visible anyway. So I experimented with a single-page design, and I came up with this.

client/components/BoatPanel/SeatingDock.js
mienaikoe: Ohh because you don't want to drag&drop from the bottom of the page if you have like 40 ppl. True.

mienaikoe: I can't promise its height will be exactly the boat, but I can estimate.

client/pages/Boat/index.js
k3ntako: if we aren't using that space for balance, do you think we should keep the boat tabs?

I know this would be undoing a lot of your work. Would it be worth it?

mienaikoe: What space?

mienaikoe:
Ooh good points. I think it would be easy to add an action to the modal that says "clear seat" or something like that.

PR #105, opened 2019-05-07
Event form bug fixes

k3ntako:
1) If user didn't click into the notes, `DraftHelpers.toRaw` would throw an error.
2) When user submits the event, they would be asked if they are sure about leaving the page.


client/components/Draft-js/DraftHelpers.js
```diff
@@ -17,6 +17,7 @@ const DraftHelpers = {
   },
 
   toRaw: function(editorState){
+    editorState = editorState ? editorState : EditorState.createEmpty();
```
mienaikoe: this can just be `editorState = editorState || EditState.createEmpty()`

client/pages/EventForm/EventForm.js
```diff
@@ -216,7 +220,7 @@ class EventForm extends Component {
       </StandIn>
 
     } else {
-      return <form className={styles.form} onSubmit={this.submitEventForm}>
+      return <form className={styles.form} onSubmit={this.onSubmit}>
```
mienaikoe: Why is there an onSubmit on the form anyway?

k3ntako: hmm yeah I'm not sure... I will remove it.

PR #107, opened 2019-05-15
External Session

mienaikoe:
- Save a list of sessions for each person in mongo
- Maintain the active list of sessions on login
- On Remove or setRole, update the session variables to reflect the person's state


mienaikoe:
Yea. I think I saw that too, but I thought it was just a fluke. I'll look into it.

server/controllers/SessionController.js
```diff
@@ -182,6 +187,17 @@ const SessionController = {
         }
       }
 
+      // remove old sessions before adding the new one
+      const sessions = await Promise.all(
+        person.session_ids.map( session_id => SessionUtils.getSession(session_id) )
+      );
+      person.session_ids = sessions.
+        filter( session => session ).
+        map( session => session.id );
```
k3ntako: I don't think the `sessions` that come from `getSession` have an `id`. 

When you first login, line 198 (two lines below this comment) will save the `session.id`. However, if you login from a different browser, the existing `session.id` will turn to `null` in the db, because of line 194 to 196. As mentioned above, in my testing the session in `map( session => session.id );` do not contain a field called `id`.

server/controllers/SessionController.js
```diff
@@ -346,6 +374,7 @@ const SessionController = {
       }else if( verification ){
         person.is_verified = true;
         person.verification = verification;
+        person.session_ids = [ req.session.id ]
```
k3ntako: would this logout all other sessions? If so, should we do this when someone updates their password?

server/controllers/SessionController.js
mienaikoe: I'll make sure to save the session_id by matching the index of person.session_ids instead of trying to get it directly from session.id

server/controllers/SessionController.js
mienaikoe: Hmm I think I wrote this on the assumption that the person would have had no sessions before, which isn't necessarily the case with verify. I'll update this code area to modify all sessions.

As for updating password from pre-auth, I think there are two reasons a person might do it:
1) Because they forgot their password
2) **Because they think there is suspicious activity**
  I think for reason 2, all sessions should be cleared when a password reset email is requested.

PR #108, opened 2019-05-18
Add Timestamp to Attendance

mienaikoe:
- Adds new fields to Event that represent attendees with a timestamp
- Adds a migration for copying data from attendees/non_attendees to attending/not_attending
- Modifies endpoints and models for dealing with the new fields
- Updates frontend to work with the new endpoint modifications


server/migrations/9_EventAttendanceTime.js
```diff
@@ -0,0 +1,32 @@
+const Event = require('../models/Event');
+
+
+module.exports = async function(){
+  console.log('========= 9_EventAttendanceTime');
+  
+  const events = await Event.find({ 
+    attending: { 
+      $size: 0
```
k3ntako: Wouldn't we want to get events where the `attending` is greater than 0? this seems like it's getting events without any responses.

We could use `$exists` to check if `attendees` field exists. We'll be removing that field in the near future, so it should return no events moving forward.

client/pages/Event/EventAttendance.js
```diff
@@ -35,16 +39,27 @@ class EventAttendance extends React.Component {
       });
     });
 
+    // find no-response members
+    const respondedIds = event.attending.
+      concat(event.not_attending).
+      map( attendance => attendance.person._id)
```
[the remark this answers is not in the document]
mienaikoe: yes, but the runtime complexity of that would be O(2n) instead of O(n) because it would loop through both lists twice instead of the once that it does here.

client/pages/Event/EventAttendance.js
```diff
@@ -35,16 +39,27 @@ class EventAttendance extends React.Component {
...
+
+    let noResponses = me.team.members.filter( member => {
+      return respondedIds.indexOf(member._id) === -1 &&
+        !me.team.coaches.find( coach => coach._id === member._id );
```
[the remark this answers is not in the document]
mienaikoe: oh true this is obsolete

server/migrations/9_EventAttendanceTime.js
mienaikoe: this is for getting events where the `attending` field hasn't been field. This is in contrast to the `attendees` field. Basically it's copying items from the attendees field to to attending field.

This migration will be run once, before the `attendees` field is removed, so it shouldn't check if the `attendees` field exists. Furthermore, the script requires the `attendees` field to exist to even work.

k3ntako: I think `$size: 0` means that `attending` exists, but the length is 0. My understanding is that it doesn't find `events` that don't have the field, `attending`. I think you need to use `$exists` to find events where `attending` field doesn't exist.

Here's what I did:
1) Ran migration
2) In MongoDB Compass, if you queried: `{ attending: {$exists: false} }`. (image 1)
  - Got all the events.
  - None of the events have an `attending` field, which means that the migration didn't work for me.
3) Queried`{ attending: {$size: 0}}` (image 2)
  - No results
  - None of the events have `attending` but `$size: 0` didn't return any events. This made me think that you should be using `$exists`, and this would explain why the migration didn't work.

Let me know if the migration worked for you. Maybe I messed up in the process. I'm having a hard time finding documentation about this.

Screen Shot 2019-05-21 at 3 17 02 PM
Screen Shot 2019-05-21 at 3 18 01 PM

mienaikoe: My data changed throughout the process of making this migration, so I can't really test it. I'll use exists then.

PR #110, opened 2019-05-21
Team email

k3ntako:
1) Team name in backend (no way for existing teams to add a name)
2) Joyride to for email page 
  - Tutorial doesn't fully close when I hit `x` in the top right. 
  - Biggest difference I noticed is that `onTourSeen` is different between the email tour and the site tour. They are defined in `Team/EventTab.js` and `Team/InspirationCard/index.js`, respectively.
3) Two methods in TeamController that aren't being used, but might be helpful later:
- `update()` - updates team `name` and `email`
- `checkEmailValid()` checks if a team `email` is valid. Most importantly, it checks that it is unique.

Not included:
1) Front-end for updating team `name` and `email`.

Question:
Email standards allow special characters (!#$%&'*+-/=?^_`{|}~), but Gmail restricts them to only periods. What should we allow? [Wikipedia - Email address](https://en.wikipedia.org/wiki/Email_address#Local-part)

To-do: 
1) Migration to add team name to existing teams.
2) Controller tests


mienaikoe:
On the email question: I think you should use letters and periods. Anything else will cause problems with gmail.

mienaikoe:
Screen Shot 2019-05-21 at 12 34 58 PM

This was a little confusing for two reasons:
1) You have to take the tour to know the email address you need to send to. The team's email address should be always visible on the email tab, and clicking it should copy to the clipboard.

2) Why does this step highlight the question mark icon? The two are unrelated.

client/pages/Team/InspirationCard/EmailTour.js
```diff
@@ -0,0 +1,81 @@
...
+
+let STEPS = [
+  {
+    target: `#${TourUtils.targets.email.COMPOSE}`,
+    title: 'Method 1: Wakebird Email Editor',
+    content: "Click here to write an email to the team without leaving Wakebird. You can choose who receive your email.",
+    disableBeacon: true,
+  },
+  {
+    target: `#${TourUtils.targets.email.HELP}`,
```
mienaikoe: see comment with screenshot

client/pages/Team/InspirationCard/EmailTour.js
```diff
@@ -0,0 +1,81 @@
...
+    } else if ([EVENTS.STEP_AFTER, EVENTS.TARGET_NOT_FOUND].includes(type)) {
+      // Keep internal state in sync with joyride
+      setStepIndex(index + (action === ACTIONS.PREV ? -1 : 1) );
+    }
+  }
+
+  const teamEmail = `${props.me.team.email}@teams.wakebird.com`;
+  const teamEmailHTML = <strong
+    key="part2"
+    onClick={() => copyToClipboard(teamEmail, props.issueSnackbar)}
```
mienaikoe: I had no indication that clicking on this link would do anything. It wasn't blue or underlined.

client/pages/Team/InspirationCard/EmailTour.js
```diff
@@ -0,0 +1,81 @@
...
+    //style={{"cursor": "grab"}} navigator.clipboard.writeText not supported on Safari/Edge.
+    >
+    {teamEmail}
+  </strong>;
+
+  STEPS[1].content = [
+    <span key="part1" style={{'userSelect': 'none'}}>{'Send an email to the team email address, '}</span>,
```
mienaikoe: You should start by introducing the idea that the team has an email address and where someone could find this email address.

client/pages/Team/InspirationCard/EmailTour.js
```diff
@@ -0,0 +1,81 @@
...
+    teamEmailHTML,
+    <span key="part3" style={{'userSelect': 'none'}}>, and Wakebird will route it to every coach and member on this team.</span>]
```
mienaikoe: The wording here seems a bit technical and I'm afraid people would get lost in the weeds reading it. Maybe it could say "and your email will be sent to everyone on the team".

client/pages/Team/InspirationCard/EmailTourCard.css
```diff
@@ -0,0 +1,3 @@
+.card{
+  background-image: url('/images/typewriter.jpg');
```
mienaikoe: Great image choice. Please resize it to 960px wide and crop it to approx 4:3 ratio. This is to reduce the amount of time it takes to load the image, and also reduce some costs of serving an image this large repeatedly.

server/utilities/TeamUtils.js
```diff
@@ -0,0 +1,54 @@
...
+      let notValid = {
+        valid: false,
+        message: ""
+      }
+
+      // Validate team._id and team email length
+      if( !team_id ){
+        notValid.message = "You must have a valid team ID";
+        return notValid;
+      }else if( !teamEmail || !teamEmail.length ){
```
mienaikoe: empty string is falsy. You don't need to also check for length

server/utilities/TeamUtils.js
```diff
@@ -0,0 +1,54 @@
...
+      // Find any other teams with the same team email
+      let teams = await Team.find({
+        $and: [
+          { email: teamEmail },
+          { _id: { $ne: team_id } } //excludes current team
+        ]
+      });
+
+      if(teams.length){
+        notValid.message = "That email address already taken";
```
mienaikoe: **is** already taken

server/utilities/TeamUtils.js
```diff
@@ -0,0 +1,54 @@
...
+      if( !team_id ){
+        notValid.message = "You must have a valid team ID";
+        return notValid;
+      }else if( !teamEmail || !teamEmail.length ){
+        notValid.message = "Team email must be at least a character long";
+        return notValid;
+      }else if( teamEmail.length > 23 ){
+        notValid.message = "Team email must be 23 characters or less";
+        return notValid;
+      }else if( !EMAILREGEX.test(teamEmail + '@teams.wakebird.com') ){
```
mienaikoe: you can simplify this if you don't include `@` in the regex or the test string. The first part of the email can be tested against the first part of the regex because you know that the rest is valid since you made it yourself.

k3ntako:
Question:
Should we even allow periods in our emails? Gmail does allow them, however, they also ignore them. My guess is that Google thinks people will confuse emails that are only a period apart.

Comment:
To answer your comment, there is no front-end for editing team name/email. As discussed earlier, changing an email is risky. And considering that this is a PR about emails, I thought changing the name would be out of the scope. I did write it, so I could include it in this PR.

server/tests/TeamControllerTest.js
```diff
@@ -274,6 +275,148 @@ describe('TeamController', function() {
...
 
+  describe(`#update`, function(){
+    before(async function() {
+      await Team.create({
+        _id: new mongoose.Types.ObjectId(),
+        name: "Team Email Test",
+        email: "coolemail",
+        // invites: [new mongoose.Types.ObjectId()]
+        // invites was added because otherwise the test would fail with following error:
+        // MongoError: E11000 duplicate key error collection: test.teams index: invites_1 dup key: { : undefined }
```
k3ntako: I keep getting an issue when I try to create a new team saying `MongoError: E11000 duplicate key error collection`

It happens when...
1) creating a team from MongoDB Compass or command line
- Fix: delete index using MongoDB Compass (not a solution)
2) running tests (line 280 - aka this comment). This specifically says duplicate key for `Team.invites`
- Fix 1: adding unique `Team.invites` array (line 284) (not a solution)
- Fix 2: removing `unique: true` from `Team.invites` in the `Team` model. 

Issue seems to be that when you make a field unique, it indexes. However, if the array is empty, it sets it to `null` (seems like it's `undefined` in our test line 286). 
https://stackoverflow.com/questions/24430220/e11000-duplicate-key-error-index-in-mongodb-mongoose

The issue isn't just that we can't have empty arrays, but why are they checking between teams? A person can be an invite on multiple teams. Should we remove `unique: true` from `Team.invites`, `members`, `coaches`, etc.?

mienaikoe: I put a commit into dev earlier today that resolves this problem. It's in migration 7. Unfortunately the documentation on mongoose-unique-array isn't the best, so I added `unique: true` thinking it would do one thing, when it would do two things. I've now removed the unique constraint in the dev branch.

mienaikoe:
Better, but consider the following:

Screen Shot 2019-05-22 at 6 19 01 PM

you're overloading the meaning of a heading by making the link the same size. Compare this to examples such as:

https://i.pinimg.com/originals/05/48/bc/0548bc477f82c5f1ef21230d04910cbc.png

client/pages/Team/EmailTab.js
```diff
@@ -130,21 +134,36 @@ class EmailTab extends Component {
...
-      <h3 className={styles.teamEmailAddress} id={TourUtils.targets.email.ADDRESS}>
-        Team Email: <a onClick={this.copyEmail}>{teamEmail}</a>
-    </h3>
-      <input id={styles.emailCopy} value={teamEmail} readOnly/>
+      <Subheading label="Team Email">
+        <div className={styles.buttons}>
+          <Button floating neutral mini
+                  icon="help"
+                  id={TourUtils.targets.email.HELP}
+                  onClick={this.startEmailTour} />
```
k3ntako: Not a big deal, but seems like you're tabbing.

PR #111, opened 2019-05-21
Boat Tweaking

mienaikoe:
- Using Eye icon instead of lock icon to show visibility
- Preventing dnd on mobile
- Using "--" to show an empty seat
- Paddler modal on mobile has every attendee, not just the ones who aren't on the boat yet. (This is because dnd was previously the only way to swap people)


client/components/BoatPanel/Seat.js
```diff
@@ -94,8 +95,8 @@ class Seat extends Component {
       seatClass.push(styles.assigned);
     } else {
       children = <div className={styles.paddler}>
-        <span className={styles.desktop}>Drag to Assign</span>
-        <span className={styles.mobile}>Tap to Assign</span>
+        <span className={styles.desktop}>--</span>
+        <span className={styles.mobile}>--</span>
```
k3ntako: issue with this is that now it's not clear how we are supposed to assign people on a user's first time. Should we have a tutorial?

client/components/BoatPanel/Seat.js
```diff
@@ -126,7 +127,7 @@ class Seat extends Component {
       seatClass.push(styles.assigned);
     } else {
       children = <div className={styles.paddler}>
-        Not Assigned
+        --
```
k3ntako: same as above.

client/components/BoatPanel/Seat.js
mienaikoe: yea. I'll add a tutorial.

mienaikoe:
- Move Tabs into Card
- Dropdowns get cut off

client/components/Subheading/index.css
```diff
@@ -9,4 +9,30 @@
...
+.collapse{
+  display: block;
+}
+
+.root.tabs h3{
+  text-align: center;
+  cursor: pointer;
+  /* font-size: 1.5rem;
+  line-height: 3rem;
+  margin: 1rem 0; */
```
k3ntako: comment

client/pages/Boat/index.js
```diff
@@ -116,6 +116,7 @@ class BoatPage extends Component {
 
       return <Tab key={boat.name} label={boat.name} icon={lock}>
         <BoatTab
+          team={me.team}
```
k3ntako: Seems like `BoatTab` doesn't need `team`.

client/pages/Boat/BoatComparison.js
```diff
@@ -0,0 +1,102 @@
...
+      events: [eventToSource(props.event)],
+      event: props.event,
+      boats: props.event.boats.map(boatToSource)
+    };
+    this.onEvent = this.onEvent.bind(this);
+    this.onBoat = this.onBoat.bind(this);
+  }
+
+  componentDidMount(){
+    Event.listForMyTeam("past").then( events => {
```
k3ntako: what if we only got the Events that have boats? Would that be confusing?

client/pages/Boat/index.js
mienaikoe: It was needed as it passed down team to the dock, which passed it to paddlers. But I can pass it directly from dock to paddlers and bypass this.

client/pages/Boat/BoatComparison.js
mienaikoe: While it's possible, I think it's a bit of a corner case. Pretty much every event coach has should have boats.

PR #112, opened 2019-05-23
Backend Refactor

mienaikoe:
- Standardized the syntax of async controller methods
- Removed extraneous findAndUpdates
- used lean wherever possible (there was maybe one spot)


server/controllers/BoatController.js
```diff
@@ -7,92 +7,108 @@ const Boat = require('../models/Boat')
...
+        seatPuller = {
+          $pull: {
+            seats: {
+              person: req.body.person_id
+            }
+          }
+        };
+      }
+    
+      if( seatPuller ){
```
k3ntako: On line 68, you define `seatPuller = {}`. Wouldn't an empty object return `true`?

server/controllers/BoatController.js
```diff
@@ -102,24 +118,24 @@ const BoatController = {
...
-      }, {
-        name: req.body.name,
-      }, {
-        new: true
-      }
-    ).then( boatMongoose => {
-      res.json(boatMongoose);
+  async update(req, res, next){
+    try{
+      const boat = await Boat.findOneAndUpdate({
```
k3ntako: could we benefit from using `updateOne` instead?

server/controllers/BoatController.js
```diff
@@ -144,34 +163,35 @@ const BoatController = {
...
-      }
-    ).then( () => {
+  async setVisible(req, res, next){
+    try{
+      await Boat.updateOne({
+          _id: req.params.boat_id
+        }, {
+          isVisible: req.body.isVisible,
+        }, {
+          new: true
```
k3ntako: no need to specify this since you're not using the returned object

server/controllers/TeamController.js
```diff
@@ -101,7 +101,7 @@ const ROLES = ['coaches', 'members', 'inactives'];
 
 const TeamController = {
 
-  mine: async function(req, res, next){
+  async mine(req, res, next){
     const isCoach = req.session.is_coach;
```
k3ntako: don't we need a `try` for `async` functions?

server/controllers/TeamController.js
```diff
@@ -117,7 +117,7 @@ const TeamController = {
     next();
   },
 
-  invitePeople: async function(req, res, next){
+  async invitePeople(req, res, next){
     let { emails } = req.body
```
k3ntako: same as above; do we need a `try`?

server/controllers/AssignmentController.js
```diff
@@ -55,14 +55,11 @@ const AssignmentController = {
     next();
   },
 
-  assign: async function(req, res, next){
-    const team = await Team.findById(req.session.team_id);
-    const assignee = await Person.findById(req.params.person_id);
-
-    await Assignment.findOneAndUpdate({
+  async assign(req, res, next){
+    await Assignment.updateOne({
```
k3ntako: same as above - don't we need a `try` ?

server/controllers/AssignmentController.js
```diff
@@ -28,24 +28,24 @@ const AssignmentController = {
...
     })
 
     res.json({ okay: true });
     next();
   },
 
-  assignMe: async function(req, res, next){
-    await Assignment.findOneAndUpdate({
+  async assignMe(req, res, next){
+    await Assignment.updateOne({
```
k3ntako: don't we need a `try` for `async` functions?

server/controllers/AssignmentController.js
```diff
@@ -28,24 +28,24 @@ const AssignmentController = {
     next();
   },
 
-  delete: async function(req, res, next){
-    await Event.findOneAndUpdate({
+  async delete(req, res, next){
+    await Event.updateOne({
```
k3ntako: don't we need a `try` for `async` functions?

server/controllers/AssignmentController.js
```diff
@@ -10,15 +10,15 @@ const Assignment = require('../models/Assignment')
 
 const AssignmentController = {
 
-  create: async function(req, res, next){
+  async create(req, res, next){
     const unsavedAssignment = new Assignment({
```
k3ntako: don't we need a `try` for `async` functions?

server/controllers/BoatController.js
mienaikoe: true. I'll not define seatPuller at 68

server/controllers/BoatController.js
mienaikoe: Yea. It looks like the frontend doesn't use the returned boat, so I'll change it to updateOne and remove the whole object return.

PR #113, opened 2019-05-23
Inbound parse

k3ntako:
To do:
1) controller tests
2) As I was writing this I found [@sendgrid/inbound-mail-parser](https://www.npmjs.com/package/@sendgrid/inbound-mail-parser). Maybe we should use this instead of multer

Question:
1) Should we include invites in emails? 
- I say yes, because as a coach it's easy keep track of who's invited, but not who made an account. It will be hard to retroactively forward them the emails they missed.

Things I learned:
1) [ngrok](https://ngrok.com/) was very helpful. When SendGrid makes the POST request to our website, we can have `localhost` respond. 
- In SendGrid's Inbound Parse, you just have to point it to the address provided by ngrok. 
- Example `https://ef9gf323.ngrok.io/api/team-emails` instead of `https://www.wakebird.com/api/team-emails`
2) Inbound parse requires a middleware in Express for handling `multipart/form-data`. 
- Couldn't get [express-formidable](https://www.npmjs.com/package/express-formidable) to work with files.
- [multer](https://www.npmjs.com/package/multer) will take the text fields and put them in `req.body` and any files will go to `req.files`
- SendGrid has an option to send "raw, full MIME message". Do not select that because multer works with `multipart/form-data`
3) If our API endpoint (`https://www.wakebird.com/api/team-emails`) returns a 400/500 status code, SendGrid seems to keep trying.
- If controller doesn't return `res.json({okay: true})` (content probably doesn't matter), ngrok perceives it as a 404 (might have been a different status code).

Helpful resources:
- [Setting Up The Inbound Parse Webhook](https://sendgrid.com/docs/for-developers/parsing-email/setting-up-the-inbound-parse-webhook/)
- [Beginner’s Technical Guide to the SendGrid Parse Webhook](https://sendgrid.com/blog/parse-webhook-tutorial/) - not very helpful but one of the few examples, and found ngrok here.
- [Sending Emails with Attachments](https://github.com/sendgrid/sendgrid-nodejs/blob/master/use-cases/attachments.md)
- [Event Webhook Node.js Code Example](https://sendgrid.com/docs/for-developers/tracking-events/nodejs-code-example/)


k3ntako:
- Return 200 for SendGrid
- File limit
- https://stackoverflow.com/questions/12416738/how-to-use-herokus-ephemeral-filesystem
- 20 MB limit

server/controllers/EmailController.js
```diff
@@ -270,6 +283,50 @@ const EmailController = {
...
+
+      const sender = await Person.findOne({ email: senderEmail });
+
+      let isMember = !!team.coaches.find(coach_id => coach_id.equals(sender._id));
+      if( !isMember ){
+        isMember = !!team.members.find(member_id => member_id.equals(sender._id));
+      }
+
+      if( !isMember ){
+        throw new ForbiddenError("You must be a coach or a member to send the emails");
```
k3ntako: invites and inactives should be able to send team emails

package.json
```diff
@@ -26,6 +26,7 @@
     "mongodb": "^3.1.10",
     "mongoose": "^5.4.0",
     "mongoose-unique-array": "^0.3.1",
+    "multer": "^1.4.1",
```
mienaikoe: Ideally, you should use libraries provided by the api creators, as they're typically going to be better-supported. You did make this work with multer though, so nbd for now.

server/controllers/EmailController.js
```diff
@@ -270,6 +283,94 @@ const EmailController = {
...
   },
 
+  routeTeamEmail: async function(req, res, next){
+    const { to, html, from, text, subject } = req.body;
+
+    let senderEmail;
+    if( from && typeof from === 'string' && from.length ){
+      const emailStart = from.indexOf('<') + 1;
+      const emailEnd = from.indexOf('>');
+      senderEmail = from.trim().toLowerCase().slice( emailStart, emailEnd );
```
mienaikoe: Doing a trim before the slice means that the indices you used to find `<` and `>` are no longer accurate. Slice the `from` field in a separate line so you can use it everywhere reliably.

server/controllers/EmailController.js
```diff
@@ -270,6 +283,94 @@ const EmailController = {
...
+          content: file.buffer.toString('base64'),
+          filename: file.originalname
+        }
+      });
+
+      const coaches = team.coaches.map(coach => coach.email);
+      const members = team.members.map(member => member.email);
+      const invites = team.invites.map(invite => invite.email);
+      const inactives = team.inactives.map(inactive => inactive.email);
+      const recipients = coaches.concat(members).concat(invites).concat(inactives);
```
mienaikoe: I think we talked about not sending emails to inactives. Can't remember where we ended up on it, but I'm personally against sending emails to inactives.

server/controllers/EmailController.js
```diff
@@ -270,6 +283,94 @@ const EmailController = {
...
 
+  routeTeamEmail: async function(req, res, next){
+    const { to, html, from, text, subject } = req.body;
+
+    let senderEmail;
+    if( from && typeof from === 'string' && from.length ){
+      const emailStart = from.indexOf('<') + 1;
+      const emailEnd = from.indexOf('>');
+      senderEmail = from.trim().toLowerCase().slice( emailStart, emailEnd );
+    }
```
mienaikoe: What if the if-condition isn't true? You should raise an exception.

server/controllers/EmailController.js
```diff
@@ -270,6 +283,94 @@ const EmailController = {
...
+      const team = await Team.findOne({ email: teamEmail }).populate(TEAM_POPULATOR);
+      if( !team ){
+        throw new ForbiddenError("The email address did not match any existing teams' email address.");
+      }
+
+      let isMember = !!team.coaches.find(coach_id => coach_id.equals(sender._id));
+      if( !isMember ){
+        isMember = !!team.members.find(member_id => member_id.equals(sender._id));
+      }
+      if( !isMember ){
```
mienaikoe: you could improve the readability and compactness of these statements by saying:
```
const isMember = 
   !!team.coaches.find(...) ||
   !!team.members.find(...) ||
...
```

server/controllers/EmailController.js
```diff
@@ -270,6 +283,94 @@ const EmailController = {
...
+        throw new ForbiddenError("The sender's email was not valid.");
+      }
+
+      const sender = await Person.findOne({ email: senderEmail });
+      if( !sender ){
+        throw new ForbiddenError("The sender's email did not match any existing users' email address.");
+      }
+
+      const teamEmail = to.trim().toLowerCase().slice(0, to.indexOf('@'));
+      const team = await Team.findOne({ email: teamEmail }).populate(TEAM_POPULATOR);
```
mienaikoe: The populator will turn your array fields into full objects instead of just ids. your `isMember` checks are expecting ids.

server/routes.js
```diff
@@ -167,6 +174,10 @@ router.route( '/emails/:email_id/include').
 router.route( '/emails/:email_id/exclude').
   post(                                                   EmailController.exclude);
 
+// Team Emails
+router.route( '/team-emails').
```
mienaikoe: make this path something a bit more generic so you can use a similar one for event attendees. Maybe like `/inbound/team` or just `/inbound` if sendgrid only allows one inbound parse endpoint.

server/tests/EmailControllerTest.js
```diff
@@ -392,4 +392,21 @@ describe('EmailController', function() {
       EmailController.deleteMany(req, res, function(){});
     })
   })
+
+  describe('#routeTeamEmail()', function() {
+
+    it('should return { okay: true } regardless of input', function(done) {
```
mienaikoe: Consider writing another test that uses Sinon.js to mock `EmailUtils.sendTeamEmail` and verify that the expected email will go to sendgrid.

https://railsware.com/blog/mocking-es6-module-import-without-dependency-injection/

server/controllers/EmailController.js
k3ntako: Yeah I think in the long term, it doesn't make sense.

I think if we make custom groups, we can have a way to include everyone.

server/tests/EmailControllerTest.js
k3ntako: I'll come back to it after i do the attendees email

server/controllers/EmailController.js
k3ntako: I need this to create the recipients list. 

Also, I'm no longer using it to check if they are a member.

server/controllers/EmailController.js
```diff
@@ -270,81 +285,88 @@ const EmailController = {
...
       const teamEmail = to.trim().toLowerCase().slice(0, to.indexOf('@'));
-      const team = await Team.findOne({ email: teamEmail });
+      const team = await Team.findOne({ email: teamEmail }).populate(TEAM_POPULATOR);
       if( !team ){
         throw new ForbiddenError("The email address did not match any existing teams' email address.");
       }
 
-      if( !team._id.equals(sender.team) ){
+      // inactives do not have a team
+      const isMember = team._id.equals(sender.team)
```
mienaikoe: This isn't great, as we're planning to remove the team field on a person. I don't want to add extra work that we need to remove later.

server/controllers/EmailController.js
```diff
@@ -270,81 +285,88 @@ const EmailController = {
...
+        || !!team.inactives.find(invite_id => invite_id.equals(sender._id));
```
mienaikoe: inactives are not invites. I suggest a variable name change.

server/controllers/EmailController.js
```diff
@@ -270,81 +285,88 @@ const EmailController = {
...
-        }
-      });
+      const attachments = InboundParseUtils.getAttachments( req.files );
 
       const coaches = team.coaches.map(coach => coach.email);
       const members = team.members.map(member => member.email);
       const invites = team.invites.map(invite => invite.email);
-      const recipients = coaches.concat(members).concat(invites);
+      let recipients = coaches.concat(members).concat(invites);
+      recipients = InboundParseUtils.removeSender( sender, recipients );
```
mienaikoe: idk.. do you need to remove the sender? If I was sending an email to the team, I'd want to be included on it so I can send replies and all that (Since the actual email that gets sent will be a totally different email thread)

server/controllers/EmailController.js
```diff
@@ -270,81 +285,88 @@ const EmailController = {
...
+      if( !event.attending.length ){
+        throw new ForbiddenError("There are no attendees for this event.");
+      }
+
+      const attachments = InboundParseUtils.getAttachments( req.files );
+
+      const coaches = team.coaches.map(coach => coach.email);
+      const attendees = event.attending.map(attendee => attendee.person.email);
+      let recipients = coaches.concat(attendees);
+      recipients = InboundParseUtils.removeSender( sender, recipients );
```
mienaikoe: see comment above.

server/controllers/EmailController.js
k3ntako: oh right, i was thinking it would be in the sent email, but it wont have the recipients

server/tests/EmailControllerTest.js
mienaikoe: I added the sinon test

PR #114, opened 2019-05-25
Copy attendees email

k3ntako:
This is draft. Just wanted to put out there my thoughts.

1) I changed the input to a textarea. The event attendees email is kinda long so it gets cut off in an input.
- Textarea can't autoresize, unless we add an resize event-listener, which might be expensive.
  - That means that it has to always be two lines tall, even if we only need one.
- Instead, I think the hidden input, with an `<h3>` works best, because headings can wrap text and dynamically change size.
2) I made the dialog bigger for bigger devices. Previously, it would go from 300px (50%) to 540px  (90%) when it hits the 600px cutoff. Now it's 400px (50%) -> (600px-450px)(75%) -> 540px (90%).
3) had to move snackbar out of `<ClipboardCopy>`, because if `<ClipboardCopy`> is in a `<Dialog>`, it's outside the `Consumers`.
4) `<Dialog>` was is basically a card, and so is `<ClipboardCopy>`. I added an option to remove the change the `<Card>` to a `<div>`
5) Textarea deselects after copying.


mienaikoe:
I tried a hidden input and the text wouldn't copy on my machine. Even when it was visible but absolutely positioned. I really don't think it's a big deal for it to be cut off.

client/components/ClipboardCopy/index.css
```diff
@@ -1,20 +1,24 @@
...
+  width: 100%;
   font-size: 1.5rem;
   font-weight: bold;
   display: block;
   text-align: center;
   border: 0;
+  outline: none;
   line-height: var(--button-height);
-  margin-bottom: 1rem;
+  margin-bottom: 0rem;
```
mienaikoe: This was necessary for the team email

client/components/ClipboardCopy/index.js
```diff
@@ -29,26 +29,31 @@ class ClipboardCopy extends React.Component {
...
-  render(){
-    return <Card className={styles.root} id={this.props.id}>
-      <input
-          className={styles.value}
-          value={this.props.value}
-          ref={ el => this._refs_input = el }
-          readOnly />
+  renderContent(){
+    return <>
+      <textarea
```
mienaikoe: i still think this should be an input. Nobody needs to see the whole thing if they have the copy button.

client/components/ClipboardCopy/index.js
```diff
@@ -29,26 +29,31 @@ class ClipboardCopy extends React.Component {
...
 
-export default function(props){
-  return <SnackbarConsumer>
-    { issueSnackbar => <ClipboardCopy {...props} issueSnackbar={issueSnackbar} /> }
-  </SnackbarConsumer>
-}
\ No newline at end of file
+  render(){
+    console.log(this.props);
+    if( this.props.showCard ){
```
mienaikoe: instead of making this an option, just use div and then require the parent to wrap this component in a card if it wants the display of a card.

client/pages/Event/AttendeesEmailDialog.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  return <Dialog key="add_person"
+      active={props.active}
+      onOverlayClick={props.onDismiss}
+      title="Email All Attendees">
+    <ClipboardCopy value={`${props.event._id}@attendees.wakebird.com`} issueSnackbar={props.issueSnackbar}>
+      <p> Any email sent to this email address will be forwarded to everyone attending this event. </p>
+      <p> The email cannot exceed 20MB in size, including attachments. </p>
+      <p> Please send messages using the email address you used to sign-up for Wakebird. </p>
```
mienaikoe: what do you think about putting all of these sentences in the same <p> tag?

client/pages/Event/EventAttendance.js
```diff
@@ -214,6 +214,7 @@ class EventAttendance extends React.Component {
             disabled={!this.state.checkedAttendees.length}>
           Mark As Not Attending
         </Button> }
+        <Button floating primary mini icon="email" onClick={this.props.openEmailDialog}/>
```
mienaikoe: personal preference: I think the button should be text+icon. Right now, it's really close to the other circle button(s) on the heading and ends up clashing a bit.
Screen Shot 2019-05-26 at 2 24 29 PM

mienaikoe: Also clashes because of alignment. I think the attendance buttons should be aligned to the right side as well, regardless of whether the email button is circle or rectangle.

client/pages/Event/AttendeesEmailDialog.js
k3ntako: Yeh i think that would look better. might be harder to read?

client/pages/Event/index.css
```diff
@@ -138,7 +141,7 @@
   }
 
   .attendeesButtons{
-    flex-direction: column-reverse;
+    /* flex-direction: column-reverse; */
   }
```
k3ntako: we can delete this

client/pages/Event/EventSidebar.js
```diff
@@ -38,18 +38,10 @@ function EventSidebar({ event, openEmailDialog }){
...
-            </a>
-          </div>
-          <div>
-            <SnackbarConsumer>
-              { issueSnackbar => <CopyButton issueSnackbar={issueSnackbar} value={`${event._id}@attendees.wakebird.com`}/> }
-            </SnackbarConsumer>
-          </div>
+          <div>Message Attendees</div>
+          <a onClick={openEmailDialog}>
+            More Information ▣
```
k3ntako: what's this box after "Information" supposed to be?

looks like this to me:
Screen Shot 2019-05-26 at 3 50 45 PM

mienaikoe: Like a modal

idk it's kind of an experiment

PR #115, opened 2019-05-27
Minor fixes

k3ntako:
1) `issueSnackbar` wasn't being passed down
2) changed API end point to `/inbound/teams` instead of `/inbound/team` to be more consistent with the email address `@teams.wakebird.com`
3) Removed custom email CSS, which was used for matching styles in Draft.js


client/pages/Team/EmailTab.js
```diff
@@ -176,9 +179,7 @@ class EmailTab extends Component {
 
 
 export default function(props){
-  return <SnackbarConsumer>
-    { issueSnackbar => <MeConsumer>
-      { meState => <EmailTab {...props} issueSnackbar={issueSnackbar} me={meState.me} /> }
-    </MeConsumer> }
-  </SnackbarConsumer>
+  return <MeConsumer>
```
mienaikoe: What's the point of moving a Consumer from the default export to the render?

server/routes.js
```diff
@@ -174,7 +174,7 @@ router.route( '/emails/:email_id/exclude').
   post(                                                   EmailController.exclude);
 
 // Inbound Emails
-router.route( '/inbound/team').
+router.route( '/inbound/teams').
```
mienaikoe: Just to be sure, you changed this URL in sendgrid as well, right?

client/pages/Team/EmailTab.js
k3ntako: It's not being used anywhere else, so I thought it was better to minimize it's scope

mienaikoe: That doesn't save any memory or processing time, and it makes the code harder to read.

Additionally, if something else in this file needed to use the consumer you would have to refactor the class.

k3ntako: I can move it out

PR #120, opened 2019-05-28
Introducing SubheadingActionable

mienaikoe:
- Supply a list of actions for a subheading to be rendered as buttons
  on desktop, or menu items on mobile.
- Add the attending setting to no-responses


client/components/Subheading/Actionable.js
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import { IconMenu, MenuItem } from 'react-toolbox/lib/menu';
+import Button from '../Button';
+import styles from './index.css';
+
+
+function getActionClass(action){
+  if( action.primary ){
+    return styles.primary;
```
k3ntako: would a `switch` be easier to read here?

client/components/Subheading/Collapsible.js
```diff
@@ -0,0 +1,18 @@
+import React, {useState} from 'react';
+import FontIcon from 'react-toolbox/lib/font_icon';
+
+
+export function CollapsibleSubheading(props){
+  const [collapsed, setCollapsed] = useState(props.defaultCollapsed || false);
+
+  return <div className={props.className || ''}>
+    <div className={`${styles.root} ${styles.collapsible}`}
```
k3ntako: you're using `styles` but it's never `import`ed.

client/components/Subheading/Collapsible.js
```diff
@@ -0,0 +1,18 @@
+import React, {useState} from 'react';
+import FontIcon from 'react-toolbox/lib/font_icon';
+
+
+export function CollapsibleSubheading(props){
```
k3ntako: did you mean `export`**`default`**?

client/components/Subheading/Actionable.js
```diff
@@ -0,0 +1,69 @@
...
+    this.setState({ isMenuOpen });
+  }
+
+  render(){
+    const { actions, label } = this.props;
+
+    const desktopActions = [];
+    const mobileActions = [];
+    actions.forEach( action => {
+      desktopActions.push(<Button key={action.label} {...action} />);
```
k3ntako: do you think it would be helpful to have a comment on the top of this file to specify what `action` should look like?
I guess it's just whatever `Button` or `MenuItem` takes?

client/pages/Event/EventAttendance.js
```diff
@@ -205,41 +205,64 @@ class EventAttendance extends React.Component {
 
   render(){
     const { me, event } = this.props
-    const { team } = me;
     const isCoach = me.isCoach();
 
+    const attendeeActions = [{
+      primary: true,
+      icon: "email",
+      onClick: this.props.openEmailDialog,
```
k3ntako: should attendees email be disabled/hidden if there are no attendees?

Should only attendees be able to email other attendees? Non-attendees won't get the replies. if so, should we disable that button here? (Also, we should probably change the back-end in a separate PR, if we are doing this).

client/pages/Inactives/index.js
```diff
@@ -94,24 +94,28 @@ class InactivePage extends React.Component {
...
-          </div>
-        </Subheading>
+        <SubheadingActionable key="members" label="Inactive Members"
+          actions={[{
+            label: "Remove",
+            accent: true,
+            flat: true,
+            icon: "remove_circle",
+            onClick: this.removeChecked,
+            disabled: this.state.checkedInactives.length <= 0
```
k3ntako: seems like `array.length` can't be less 0. You could just test for `... checkedInactives.length === 0`. 

https://stackoverflow.com/questions/32386253/can-the-js-array-length-property-ever-return-a-negative-value

client/pages/Inactives/index.js
```diff
@@ -94,24 +94,28 @@ class InactivePage extends React.Component {
...
+          },{
+            label: "Invite to Team",
+            primary: true,
+            flat: true,
+            icon: "person_add",
+            onClick: this.inviteChecked,
+            disabled: this.state.checkedInactives.length <= 0
```
k3ntako: same as above

client/components/Subheading/Actionable.js
mienaikoe: I was trying to figure out how to do a switch here. The problem is that the API for react-toolbox accepts a bunch of boolean parameters instead of a string, and switches only deal with single variables.

client/pages/Event/EventAttendance.js
mienaikoe: Disabled the button, and hiding the button for non-attendees or non-coaches

mienaikoe: Also added a trello card

PR #121, opened 2019-05-29
Gravatar roster

k3ntako:
Still to do:
- ~show image on person's page~
- ~instructions on how to add profile photo on Gravatar~


client/pages/Team/Roster.js
```diff
@@ -82,7 +82,21 @@ class Roster extends React.Component {
       birthday = EventFormatter.formatBirthdayShort(person.birthday);
     }
 
+    const style = {
+      backgroundColor: person.backgroundColor,
+      borderColor: person.backgroundColor,
```
mienaikoe: I wonder if it would look nice if the border was either a little lighter or a little darker, and like 4 pixels thick

client/pages/Team/Roster.js
```diff
@@ -82,7 +82,21 @@ class Roster extends React.Component {
...
+    }
+
     const cells = [
+      <TableCell key="picture">
+        <Link to={`/person/${person._id}`}>
```
mienaikoe: Let's make this snippet of jsx into a reusable component

mienaikoe: from Link down to the end of Link

client/components/ProfilePicture/large.css
```diff
@@ -0,0 +1,3 @@
+.picture{
+  border-radius: 10px;
+}
```
mienaikoe: This file is largely not necessary. You could just put all the styles in one file. Especially if you plan to use styles common to both.

client/components/ProfilePicture/small.css
```diff
@@ -0,0 +1,20 @@
+.profile{
+  height: 50px;
+  width: 50px;
+  border: 4px solid;
+  border-radius: 10px;
```
mienaikoe: Profile images are always circles. Github is the only place i've seen a square profile image. Every other site in the world gives a circle.

client/components/ProfilePicture/small.css
```diff
@@ -0,0 +1,20 @@
...
+  display: flex;
```
mienaikoe: flex isn't really necessary unless the image isn't square. Does gravatar not guarantee squareness in its avatars?

client/components/ProfilePicture/small.css
```diff
@@ -0,0 +1,20 @@
...
+  justify-content: center;
+}
+
+.profile object {
+  position: relative;
+  top: -4px;
+  height: 50px;
+  display: flex;
+  align-items: center;
```
mienaikoe: again, not really necessary unless the avatar isn't square.

client/models/Person.js
```diff
@@ -139,8 +140,17 @@ class Person extends Model {
   }
 
   get backgroundColor(){
-    let hue = ((this.first_name.charCodeAt(0) + this.last_name.charCodeAt(0)) * 10) % 255;
-    return `hsl(${hue}, 90%, 50%)`;
+    let hue = ((this.first_name.charCodeAt(0) + this.last_name.charCodeAt(0)) * 10) % 360;
+    return `hsl(${hue}, 90%, 60%)`;
```
mienaikoe: I still worry about the yellow being too bright. Maybe there can be a slight dip in lightness around the yellow hue. 

I think the initials color might need to be black or white depending on the background color (black for yellow, and white for all else).

client/models/Person.js
```diff
@@ -139,8 +140,17 @@ class Person extends Model {
...
+  }
+
+  get borderColor(){
+    let hue = ((this.first_name.charCodeAt(0) + this.last_name.charCodeAt(0)) * 10) % 360;
```
mienaikoe: If this duplicates the calculation of backgroundColor, then maybe it's best to just have a function for "hue", and let the caller create the hsl string on its own, as it will likely be design-dependent.

client/pages/Me/index.css
```diff
@@ -93,6 +93,32 @@
   margin-right: 2rem;
 }
 
+.profilePicture{
+  display: flex;
+}
+.profilePicture{
```
mienaikoe: duplicate rule

client/pages/Me/index.css
```diff
@@ -93,6 +93,32 @@
...
+  max-width: 200px;
+  margin: 15px 2rem 15px auto;
+}
+
+.instruction p,
+.instruction ol{
+  font-size: 1.3rem;
+}
+.instruction ul{
+  padding-inline-start: 1.5rem;
```
mienaikoe: what does this rule do, and why is it necessary?

client/pages/Person/Roles.js
```diff
@@ -0,0 +1,103 @@
+import React, {Component} from 'react';
```
mienaikoe: Great idea making this file.

mienaikoe:
I don't see the avatar on the person page. Is it supposed to be there?

client/models/Person.js
mienaikoe: You can also use a border color like `rgba(0,0,0,0.2)` to provide some darkness to any color, instead of having to recalculate the hue.

client/models/Person.js
mienaikoe: Actually, the initials are all black, and its fine caus the colors are pastel.

client/components/ProfilePicture/small.css
mienaikoe: Additionally, if you're 100% certain on square, you shouldn't round the corners.

client/pages/Me/MePage.js
```diff
@@ -35,6 +36,8 @@ export default class MePage extends Component {
           <UpdateMe
             me={me}
             onSubmit={this.onSubmit}/>
+          <h2> Profile Picture </h2>
+          <ProfilePicture me={me}/>
```
mienaikoe: Profile Picture should be at the top, above UpdateMe. People latch onto photos immediately, and end up trusting a site more because of it.

client/pages/Me/ProfilePicture.js
```diff
@@ -0,0 +1,39 @@
...
+import Button from '../../components/Button';
+import ProfilePictureLarge from '../../components/ProfilePicture/Large';
+import styles from './index.css';
+
+
+export default (props) => {
+  return <div className={styles.profilePicture}>
+    <ProfilePictureLarge person={props.me} />
+    <Card className={styles.form}>
+      <CardText className={styles.instruction}>
```
mienaikoe: I think these instructions could be hidden in a modal triggered from a button like "Change Photo".

Having this much text in the field makes this whole page look overwhelmingly complex. Hiding text walls and reducing lines on a page helps to clear the cognitive overload and help people understand a page better.

mienaikoe: To be honest, I kind of want to do that with the notice at the top of Personal Information too. Like put it in a modal triggered from a `?` button on the right of the subheading.

client/components/ProfilePicture/Small.js
```diff
@@ -0,0 +1,20 @@
...
+  const person = props.person;
+
+  const style = {
+    backgroundColor: person.backgroundColor,
+    borderColor: person.borderColor,
+  }
+
+  return <Link to={`/person/${person._id}`}>
+    <div className={styles.profile} style={style}>
+      <object data={person.gravatar("50")} type="image/jpeg">
```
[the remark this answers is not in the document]
mienaikoe: Oh dude there are actually a bunch of alternatives you can use:

https://en.gravatar.com/site/implement/images/

`blank` could be used for putting it immediately above the initials/color combo. This would reduce the need for two calls for every avatar.

client/pages/Person/index.css
```diff
@@ -39,33 +53,75 @@
...
 @media screen and (max-width: 720px){
   section.section{
     display: block;
     padding: 1rem;
   }
 
-  .inputRow{
+  .section h2{
+    margin-left: 2rem;
+    margin-right: 2rem;
```
mienaikoe: I don't know if this is the rule that's causing it, but there's a breakpoint at which the alignment of the heading isn't the same as the alignment of the subheadings.

Screen Shot 2019-05-30 at 12 59 30 PM

The margin above 600px is 4rem I think. Below this is 1 or 2 rem. If you want to change your margin, I think you should use the same 600px breakpoint.

client/pages/Me/index.css
k3ntako: this reduces the indentation for `<li>` inside the `<ul>`.

client/components/ProfilePicture/Small.js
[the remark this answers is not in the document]
k3ntako: oh yeah, i don't know why i didnt choose one of those...

maybe i just wanted to try out your colors. 

> blank could be used for putting it immediately above the initials/color combo.

is there a point in have colors or initials if we have a default?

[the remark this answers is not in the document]
mienaikoe: yea if the default is transparent, then you'll need the colors or initials to be the default.

client/pages/Me/index.css
mienaikoe: ooh TIL

client/pages/Me/MePage.js
k3ntako: what if they dont have a photo? Should i put a default that they send us, or nothing at all?

client/pages/Me/ProfilePicture.js
k3ntako: made a Trello ticket for notice

client/models/Person.js
mienaikoe: well considering the function relies on luminance, and our code hard-codes the luminance, I don't think it's particularly important to do this. I think I commented this elsewhere, but the color is light enough that you can probably just use black across the board. The black/white calculation would only be necessary if we introduced darker backgrounds, which I don't anticipate.

k3ntako:
Change rating to PG

client/models/Person.js
k3ntako: Here's what it looks like: https://codepen.io/K3ntako/pen/qGvWbm?editors=0010
what do you think?

I think the purples might be too dark.
By default, there is no connection between the initials and color.
Ignore the font size, because it's not consistent with what shows up on our site.

mienaikoe: I think it looks great

PR #127, opened 2019-06-09
 Assignments Redesign

mienaikoe:
- Changes person autocomplete to attendee list when assigning a person
- Adds a sidebar to assignments tab for coach only (shows who had assignments in past events)
- Refactors `SeatingDock` to a more generic `AttendeeSelect` so both `BoatPanel` and `PastAssignments` can use it


client/pages/Event/AssignmentTable.js
```diff
@@ -83,7 +82,7 @@ class AssignmentTable extends React.Component {
       <SubheadingActionable 
         label={assignment.name}
         actions={actions} />
```
k3ntako: I was talking to Yoanna yesterday, and she was saying how we should have 4 people unwrapping boats, not 2. When we have two boats, vets and rookies can do it together. I was just thinking there is no easy way of changing all the limits. 

I was also wondering if we could add a way to increase the limit from the assignments tab. For example we have a way to increase/decrease the limit next in the `< SubheadingActionable>`, next to "Assign Me". At least this would be easier than going in to Edit Event.

Any thoughts? Maybe we just need to wait for the Series Edit.

client/pages/Event/PastAssignments.js
```diff
@@ -0,0 +1,91 @@
...
+    });
+  }
+
+  onEvent( eventId ){
+    if( !this.state.event || this.state.event._id !== eventId ){
+      this.setState({
+        event: this.state.baseEvents.find( event => event._id === eventId ),
+      }, () => {
+        Event.find(eventId).then( newEvent => {
+          this.setState({
```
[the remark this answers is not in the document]
mienaikoe: To populate the assignments.

client/pages/Event/AssignmentTable.js
mienaikoe: I think this should be in the series edit. Putting forms where they shouldn't be would make the whole site impossible to understand.

client/pages/Event/PastAssignments.js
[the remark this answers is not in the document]
mienaikoe: Maybe we shouldn't remove the assignments field in the list call, and show if you have an assignment in the upcoming field as a way to bubble it up.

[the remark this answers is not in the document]
k3ntako: Yeah I think it would be best that we don't slow down the other parts of the website.

client/pages/Event/EventAssignments.js
```diff
@@ -1,25 +1,50 @@
...
+import { MeConsumer } from '../../contexts/MeContext';
+import Subheading from '../../components/Subheading';
+import styles from './EventAssignments.css';
 
 
 class EventAssignments extends React.Component {
 
+  onAssign = () => {
+    this.forceUpdate();
+  }
```
k3ntako: Seems like this function is why you added the `@babel/plugin-proposal-class-properties` library. 

We could change this to not ES6...
```
  onAssign() {
    this.forceUpdate();
  }
```
...and the one time you use it (line 26), `bind(this)`: 
```
let assignmentTables = event.assignments.map( assignment => {
  return <AssignmentTable key={assignment._id}
    assignment={assignment}
    event={event}
    onAssign={this.onAssign.bind(this)} />
});
```

I'm guessing less libraries is generally better.

k3ntako: if you do uninstall it, please remove it from `.bablerc` as well.

mienaikoe: I added this library because I was tired of binding to this everywhere. This library lets you get around binding to this because it can understand class properties and transpile them to automatically bind to `this`. In the future, I would like to remove all bindings to `this` because it's extraneous boilerplate that isn't necessary.

PR #130, opened 2019-06-11
Redux - Snackbar

mienaikoe:
- Replaces SnackbarContext with SnackbarReducer
- Creates a SnackbarRedux Component that responds to all snackbar redux state changes
- Refactors all callers of SnackbarContext


client/components/SnackbarRedux/index.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { connect } from "react-redux";
+import { Snackbar } from 'react-toolbox/lib/snackbar';
+import styles from './index.css';
+import SnackbarReducer from '../../redux/reducers/SnackbarReducer';
+
+
+const SnackbarRedux = (props) => {
+  return <Snackbar
+    className={styles[props.messageType]}
```
k3ntako: `messageType` doesn't seem to be working. When I invite someone, `messageType` is set to `true`. Normally, inviting someone would yield a green snackbar. It's the default black right now.

PR #131, opened 2019-06-12
Test for 'future', 'past' and 'all' events

k3ntako:
Make sure that EventController.listForMyTeam() returns only the events queried.


server/tests/EventControllerTest.js
```diff
@@ -165,17 +165,74 @@ describe('EventController', function() {
...
+              team: req.session.team_id,
+              datetime: {
+                $gt: midnightLastYear,
+              },
+            });
+
+            expect(response).to.be.an('array');
+            expect(response.length).to.eql(allEvents.length);
+            expect(response[0].constructor).to.eql(Event);
+            expect(response[0].datetime.getTime()).to.be.below(new Date().getTime());
```
mienaikoe: You could go through both lists and ensure all the ids are the same instead of just checking the first value

server/tests/EventControllerTest.js
```diff
@@ -165,17 +165,74 @@ describe('EventController', function() {
   })
 
   describe('#listForMyTeam()', function() {
+    let midnightThisMorning, midnightLastYear;
+    before( async function(){
+      // Less than a year old
+      await TestHelper.createEvent( coach_session, new Date().getTime() - 1000000000 );
```
mienaikoe: Might be more useful to make events that have timestamps that are only minutes before or after the cutoff point of each query so as to ensure the method is precise as well as accurate.

PR #133, opened 2019-06-17
Fix assignments behavior

k3ntako:
If all attendees were calculated `event.attendees.length <= assignment.assignees.length` within each `<AssignmentAction>` so its siblings weren't aware if the limit was reached. Something like the following would happen.

Also, added coaches to the calculation. 

Screen Shot 2019-06-16 at 8 16 57 PM


client/pages/Event/AssignmentTable.js
```diff
@@ -57,14 +57,17 @@ class AssignmentTable extends React.Component {
         assigneeLink = "-"
       }
 
+      const assignedAll = event.attendees.length + me.team.coaches.length <= assignment.assignees.length;
```
mienaikoe: Why did you move this calculation to the parent instead of the child? The parent has no use for this prop other than to pass it to the child.

k3ntako: Let's say 5 people are attending (including coaches). After the fifth person is assigned, only the `<AssignmentAction>` that was clicked with redo their calculations. That means that most of them would still say `Assign`, because `props` hasn't changed.

mienaikoe: ah makes sense. why not run this calculation outside the for loop then? It doesn't need to be recalculated for every assignee

PR #136, opened 2019-08-01
Set Team page for those with more than one team

mienaikoe:
This adds a page for setting a team. It should only be visible to those with more than one team. It also adds a switch team link to the user dropdown when you have more than one team.


client/SessionReRouter.js
```diff
@@ -81,6 +92,8 @@ class SessionReRouter extends React.Component {
       default:
         if( !this.props.me ){
           this.props.history.replace('/login');
+        } else if( !this.props.team ){
```
k3ntako: I think we should put this after the `if` block for `!this.props.me.is_verified`

client/components/Nav/index.js
```diff
@@ -66,6 +73,14 @@ class Nav extends Component {
         tooltipShowOnClick={true}
         tooltipPosition="bottom" />;
 
+    let switchTeams;
+    if( me && me.teams.length > 1 ){
+      switchTeams = <MenuItem
+          caption='Switch Team'
+          icon="swap_horiz"
+      onClick={this.switchTeam}/>
```
k3ntako: NIT: indentation for `onClick`

mienaikoe: oops! thx

PR #137, opened 2019-09-08
Birthday object

k3ntako:


server/migrations/13_BirthdayObject.js
```diff
@@ -0,0 +1,26 @@
...
+      const month = person.birthday.getUTCMonth() + 1;
+      const year = person.birthday.getUTCFullYear();
+
+      const birthdayObject = {
+        date,
+        month,
+        year,
+      };
+
+      person.birthday_object = birthdayObject;
```
mienaikoe: ok. I know this might be annoying, but typing `birthday_object` all the time seems annoying too caus it's long and two words. What about like... `birthdate`

mienaikoe:
Hmm this is marked as a draft. Is that still the case?

k3ntako:
Oops, I just opened the PR, thanks!

PR #138, opened 2019-09-23
Create new team

k3ntako:


k3ntako:
I'm wondering if we shouldn't ask for a team email on create. I feel like that could slow people down, and we want people moving past that quickly?

server/controllers/TeamController.js
```diff
@@ -71,6 +73,49 @@ const TeamController = {
     }
   },
 
+  async create(req, res, next){
+    try{
+      let { teamName, teamEmail } = req.body;
```
mienaikoe: can we rename these to `name`, and `email`? We know they're for the team already by virtue of using this endpoint.

server/controllers/TeamController.js
```diff
@@ -71,6 +73,49 @@ const TeamController = {
...
+
+      teamName = teamName.trim();
```
mienaikoe: check for existence before calling trim. Something like `name = name && name.trim()`

server/controllers/TeamController.js
```diff
@@ -71,6 +73,49 @@ const TeamController = {
...
+      if ( !teamName ){
+        throw new ForbiddenError("You must provide a team name");
+      }
+
+      // Validate team email
+      teamEmail = teamEmail.trim().toLowerCase(); // team email is stored as lowercase
```
mienaikoe: same here. `email = email && email.trim().toLowerCase()`

server/controllers/TeamController.js
```diff
@@ -71,6 +73,49 @@ const TeamController = {
...
+      res.json(newTeam);
+      next();
+    }catch( err ){
+      next(err);
+    }
+  },
+
   async update(req, res, next){
     try {
       const team = await Team.findOne({
```
mienaikoe: This findOne call should be done after validations

server/controllers/TeamController.js
```diff
@@ -115,12 +165,21 @@ const TeamController = {
...
   },
 
-  async checkEmailValid(req, res, next){
-    let { teamEmail } = req.body;
-    const isEmailValid = await TeamUtils.checkEmailValid(req.session.team_id, teamEmail);
+  async validateTeamEmail(req, res, next){
+    try{
+      const { teamEmail } = req.body;
+
+      if ( !req.session.team_id ){
```
mienaikoe: `validateTeamEmail` might be called by someone who hasn't made a team yet. Remove this check.

client/models/Team.js
```diff
@@ -10,13 +10,6 @@ class Team extends Model {
     super(props);
   }
 
-  static create( teamName, teamEmail ){
-    return Fetcher.post('/api/team/create', {
-      teamName,
-      teamEmail,
-    }).then( teamJson => new Team(teamJson));
-  }
-
```
mienaikoe: Ah yea. I had realized we both wrote it, but it was late by the time I realized it and I was sleepy lol.

PR #140, opened 2019-09-29
Create Team

mienaikoe:
The frontend, and several changes that fix ui bugs found in manual testing and automatic testing


client/Routes.js
```diff
@@ -1,48 +1,51 @@
...
-import PersonPage from './pages/Person';
-import EmailPage from './pages/Email';
-import InactivesPage from './pages/Inactives';
-import NotFoundPage from './pages/NotFound';
-
+import TeamFormPage from './pages/TeamForm';
+import TeamNonePage from './pages/TeamNone';
+import TeamPage from "./pages/Team";
+import UnverifiedPage from './pages/Auth/Unverified';
+import VerifyEmailPage from './pages/Auth/VerifyEmail';
```
mienaikoe: Just Alphabetizing some things

client/SessionReRouter.js
```diff
@@ -71,29 +73,48 @@ class SessionReRouter extends React.Component {
...
           this.props.history.replace('/home');
         }
         break;
+      case '/no-teams':
+      case '/create-team':
+        if( !this.props.me ){
+          this.props.history.replace('/login');
+        } else if( !this.props.me.is_verified ){
+          this.props.history.replace('/unverified');
+        } else if( this.props.me.teams.length && !this.props.team ){
```
mienaikoe: this is where you mentioned combining the `if` clauses before

client/pages/Auth/SetTeam/styles.css
```diff
@@ -11,12 +11,13 @@
...
   z-index: -1;
   background-color: white;
-  min-height: 80vh;
 }
 
 .teamSelect{
   width: 100%;
   max-width: 480px;
+  overflow-y: scroll;
+  height: 70vh;
```
mienaikoe: overflow issues when you make a bunch of teams

server/tests/TeamControllerTest.js
```diff
@@ -75,18 +78,21 @@ describe('TeamController', function() {
...
           try{
+            let response = JSON.parse(JSON.stringify(responseDoc));
+            let expectedCoachId = String(second_member_session.person_id);
+
             expect(response.name).to.equal(req.body.name);
             expect(response.email).to.equal(req.body.email);
-            expect(response.coaches).to.include(second_member_session.person_id);
-            expect(String(response.owner)).to.equal(String(second_member_session.person_id));
+            expect(response.coaches[0]._id).to.eql(expectedCoachId);
+            expect(response.owner._id).to.equal(String(expectedCoachId));
```
mienaikoe: Just ensuring that the team returned is a populated team. This is important because we use a bunch of deep fields from the assigned session team in the frontend. (The list of coaches and members, for instance, are used in the roster)

client/pages/TeamForm/TeamFormPage.js
```diff
@@ -0,0 +1,93 @@
...
+
+  function submit(){
+    if( isCreating ){
+      Team.create({name, email}).then( team => {
+        props.setTeam(team);
+        props.history.push('/home');
+      });
+    } else {
+      props.team.update({name, email}).then( () => {
+        cancel();
```
k3ntako: NIT: if youre going to use `cancel()` on a successful call, maybe it's better call it something like `goBack` or `goBackAPage`.

client/pages/TeamForm/TeamFormPage.js
```diff
@@ -0,0 +1,93 @@
...
+      })
+    }
+  }
+
+  function setNameSafe(inputName){
+    setName(inputName);
+    if( !isEmailSpecified ){
+      let inferredEmail = inputName.replace(/\W/g, '').toLowerCase();
```
k3ntako: This allows for `_` to pass through, which is not allowed by us right now. That is defined by `TEAMEMAILREGEX` in `server/utilities/TeamUtils`.

client/pages/TeamForm/styles.css
```diff
@@ -0,0 +1,37 @@
...
+.inputs input {
+  width: 100%;
+}
+
+.buttons{
+  text-align: right;
+  margin-top: 1rem;
+}
+
+@media screen and (max-width: 600px){
```
k3ntako: I think we need to add some padding or margin on `.container` or `.root` for mobile view. 
(iPhone 6/7/8 Plus)
Screen Shot 2019-09-30 at 11 28 20 AM

server/controllers/SessionController.js
```diff
@@ -296,12 +296,21 @@ const SessionController = {
 
         req.session.team_id = intendedTeam._id;
         req.session.is_coach = !!intendedTeam.coaches.find(coach => coach._id.equals(person._id) );
+
+        const populatedTeam = await Team.
+          findById(intendedTeam._id).
+          populate( Team.getPopulator(true) );
+
+        res.json(populatedTeam.toJSON({
+          isCoach: true
```
k3ntako: why is `isCoach` set to `true`? On line 298, you are checking if the user is a coach:
```
req.session.is_coach = !!intendedTeam.coaches.find(coach => coach._id.equals(person._id) );
```

server/tests/TeamControllerTest.js
```diff
@@ -42,18 +42,21 @@ describe('TeamController', function() {
       }
 
       const res = {
-        json: async function(response){
+        json: async function(responseDoc){
           try{
+            let response = JSON.parse(JSON.stringify(responseDoc));
```
k3ntako: why do you do this?

client/components/Nav/index.js
```diff
@@ -105,13 +109,18 @@ class Nav extends Component {
             <MenuItem
                 caption='Edit Account'
                 icon="edit"
-                onClick={this.linkToMe}/>
+                onClick={this.linkToMe.bind(this)}/>
```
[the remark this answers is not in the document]
mienaikoe: Ah. I think it was a muscle memory thing. binding to `this` seems to be unnecessary, as expected. I'll remove it.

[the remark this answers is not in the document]
mienaikoe: actually it is necessary, but the constructor is the right palce to do it. I had thought we added a webpack option for this but apparently not. Will resolve separately. For now, let's bind to `this` in the constructor or use arrow functions... or honestly bind(this) is fine too. There's no wrong way to do it because they all basically do the same thing.

client/pages/TeamForm/TeamFormPage.js
mienaikoe: man that regex is straight up insane. Can we just demand that the email have letters and numbers only?

client/pages/TeamForm/styles.css
mienaikoe: oof RIP lol

server/tests/TeamControllerTest.js
mienaikoe: This ensures that the response is just as it would be if it were coming from a `fetch` call. More specifically, it ensures that any models and their functions that were encapsulating an object no longer exist. This is especially important when comparing `_id` because those are strings after the parse/stringify, but they are ObjectId before.

mienaikoe: Ideally our tests wouldn't need to do all this boilerplate code. I think that one ticket with rewriting backend tests using `chai-http` would solve all of that for us.

server/controllers/SessionController.js
mienaikoe: oh. Goodness... I think that was a copypaste. Good eye!

client/pages/TeamForm/TeamFormPage.js
k3ntako: yeah i think that would be better. I made a new trello card for that earlier, but we can change it here instead.

One thing to note is that it allows periods currently. but the period can't be the last character. I believe that is an email requirement.

server/tests/TeamControllerTest.js
k3ntako: ahh got it! thanks for the explanation!

PR #142, opened 2019-10-02
Sequelize

k3ntako:
### Before you start:
- `npm i`
- `npx sequelize-cli db:create`  to create `wakebird_development` db.
- `npx sequelize-cli db:migrate` to migration `Sessions` table.
- `npx sequelize-cli --help` will bring up all the `sequelize-cli` commands.

#### Added Sequelize packages:
- **sequelize** - ORM for Node.js and works with different databases including Postgres.
- **sequelize-cli**: required for [migrations](https://sequelize.org/master/manual/migrations.html). Allows us to run commands from command line.
- **pg** and **pg-hstore** - specifically for Postgres.

#### `.sequelizerc` (generated and edited)
- Configs for where the Sequelize files will live
- `require("@babel-register")` [allows us to write migrations and seeds in ES6/ES7 syntax](https://sequelize.org/master/manual/migrations.html#using-babel).

#### `server/setup/postgresConfig.json` (generated and edited)
- Contains config for dev, test, and production dbs.

#### `server/sequelizeModels/index.js` (generated)
- Sets up Sequelize models and exposes them.

#### Converted Sessions to Postgres
- Migration and model for `Sessions` table (generated and edited)
- `server/setup/session.js` - changed to use Postgres. 
- `connect-session-sequelize` replaces `connect-mongo`. It's specifically built for Sequelize. Concern is that it hasn't seen an update since last Nov.


.sequelizerc
```diff
@@ -0,0 +1,10 @@
+require("@babel/register");
+
+const path = require('path');
+
+module.exports = {
+  'config': path.resolve('server/setup', 'postgresConfig.json'),
+  'models-path': path.resolve('server', 'sequelizeModels'),
+  'seeders-path': path.resolve('server', 'sequelizeSeeders'),
+  'migrations-path': path.resolve('server', 'sequelizeMigrations')
```
mienaikoe: can we use deeper directory nesting?

i.e. `server/sequelize/models` `server/sequelize/seeders`, etc

server/sequelizeMigrations/20191002010918-create-sessions.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
```
mienaikoe: was this file auto-generated?

server/setup/session.js
```diff
@@ -1,29 +1,25 @@
-const mongoose = require('mongoose')
+var Sequelize = require('sequelize');
+const { sequelize } = require('../sequelizeModels');
 const session = require('express-session');
-const MongoSessionStore = require('connect-mongo')(session);
-
-const environment = process.env['NODE_ENV']
-const config = require('../config')[environment];
-
 const SessionUtils = require('../utilities/SessionUtils');
```
mienaikoe: not related to your work, but this looks like an unused import

server/sequelizeModels/index.js
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+Object.keys(db).forEach(modelName => {
+  if (db[modelName].associate) {
+    db[modelName].associate(db);
+  }
+});
+
+db.sequelize = sequelize;
+db.Sequelize = Sequelize;
```
mienaikoe: what is the difference between lowercase sequelize and uppercase? I feel like one or both of these should be named something more descriptive.

server/sequelizeModels/index.js
```diff
@@ -0,0 +1,37 @@
...
+const path = require('path');
+const Sequelize = require('sequelize');
+const basename = path.basename(__filename);
+const env = process.env.NODE_ENV || 'development';
+const config = require(__dirname + '/../setup/postgresConfig.json')[env];
+const db = {};
+
+let sequelize;
+if (config.use_env_variable) {
+  sequelize = new Sequelize(process.env[config.use_env_variable], config);
```
mienaikoe: none of the configs you've specified in `postgresConfig.json` has `use_env_variable`

mienaikoe:
Something I noticed is that every call now includes a postgres call. Can we memoize the session request, or is that not possible out of the box?

server/sequelizeMigrations/20191002010918-create-sessions.js
k3ntako: yep, I wasn't really clear, but when i mentioned "generated" i meant auto generated. So, a few of the files in this PR were auto-generated and then edited.

server/setup/session.js
k3ntako: I think we're using at line 21: `SessionUtils.initialize(sessionStore);`

server/sequelizeModels/index.js
k3ntako: `Sequelize` (uppercase) is the library/class, and `sequelize` (lowercase) is the instance or the active connection to the library.

I'm not really sure why we need to have library on `db`. Do you think removing that would help reduce confusion?

k3ntako:
> Something I noticed is that every call now includes a postgres call. Can we memoize the session request, or is that not possible out of the box?

It doesn't look like there is anything a way to memoize out of the box. But, I'm not super familiar with this stuff, so it's possible i missed it.

server/sequelizeModels/index.js
k3ntako: I think we will end up using this later on when we set up production.

server/sequelizeMigrations/20191002010918-create-sessions.js
```diff
@@ -0,0 +1,32 @@
...
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER
+      },
+      sid: {
+        allowNull: false,
+        type: Sequelize.STRING,
+      },
+      userId: { type: Sequelize.STRING },
```
k3ntako: don't add yet

server/sequelizeModels/session.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+module.exports = (sequelize, DataTypes) => {
+  const Session = sequelize.define('Session', {
+    sid: DataTypes.STRING,
+    userId: DataTypes.STRING,
```
mienaikoe: This should be added later when we add Person to postgres migrations

server/sequelizeModels/session.js
```diff
@@ -0,0 +1,15 @@
...
+    expires: DataTypes.DATE,
+    data: DataTypes.STRING(50000),
```
mienaikoe: Make this 5,000 instead of 50,000

k3ntako:
- package.json - script for db create for test and dev
- use `NODE_ENV=test npx sequelize-cli db:create` to create 
- package.json - script for migration (put it with mongo migration)
- test: clear data but keep all the table
- test: migrate before tests

server/sequelizeModels/index.js
k3ntako: the library that we are importing, `const Sequelize = require('sequelize');`

server/sequelizeMigrations/20191002010918-create-sessions.js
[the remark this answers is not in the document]
k3ntako: I'm going to remove this for now

PR #144, opened 2019-10-08
Sequelize Person

mienaikoe:
Includes the standard code changes plus some refactoring of various things I saw.

Note: `PersonController.listForMyTeam` needs to be written when Team and Team controller are built.


server/sequelize/migrations/20191007223203-create-person.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('people', {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER
```
k3ntako: since we use our `person_id` as part of the url, should we use UUIDs?

Sequelize has data types for [UUID](https://sequelize.org/master/class/lib/data-types.js~UUID.html), [UUIDV1](https://sequelize.org/master/class/lib/data-types.js~UUIDV1.html), [UUID4](https://sequelize.org/master/class/lib/data-types.js~UUIDV4.html). The documentation for UUID says "Use with UUIDV1 or UUIDV4 for default values." I wonder if it auto-generates the UUID for us if we set it as default.

If not, the [uuid npm package](https://www.npmjs.com/package/uuid) mentions a super short [function](https://gist.github.com/jed/982883) for implementing UUID.

server/sequelize/migrations/20191007223203-create-person.js
```diff
@@ -0,0 +1,58 @@
...
+      },
+      first_name: {
+        allowNull: false,
+        type: Sequelize.STRING(500)
```
k3ntako: Should we set a limit for first names that we enforce on the front-end? I think 99.99% of people have a first name shorter than 25 letters, or have a shorter name they go by. This also applies for **last name** and **email**.

Just in case you were curious about the longest name:
https://en.wikipedia.org/wiki/Hubert_Blaine_Wolfeschlegelsteinhausenbergerdorff_Sr.

server/sequelize/migrations/20191007223203-create-person.js
```diff
@@ -0,0 +1,58 @@
...
+        type: Sequelize.STRING(20)
+      },
+      birthdate: {
+        type: Sequelize.DATEONLY
+      },
+      sex: {
+        type: Sequelize.ENUM('M', 'F')
+      },
+      location: {
+        type: Sequelize.STRING(500)
```
k3ntako: I'm thinking that a location that's 500 characters long wouldn't fit on the front-end. Unless we are future-proofing, I think we could lower the limit.

server/sequelize/models/Person.js
```diff
@@ -0,0 +1,62 @@
...
+    birthdate: {
+      type:         DataTypes.DATEONLY,
+      get() {
+        birthdate = this.getDataValue('birthdate');
+        return {
+          date: birthdate.getUTCDate(),
+          month: birthdate.getUTCMonth(),
+          year: birthdate.getUTCFullYear(),
+        };
+      },
```
k3ntako: shouldn't this be an object with numbers instead of a date?
In Mongo it looked like this:
```
birthdate: {
  date:               { type: Number },
  month:              { type: Number },
  year:               { type: Number },
},
```

server/sequelize/models/Person.js
```diff
@@ -0,0 +1,62 @@
+'use strict';
+const Sequelize = require('sequelize');
+
+module.exports = (sequelize, DataTypes) => {
+  const Person = sequelize.define('Person', {
+    first_name:     DataTypes.STRING(500),
+    last_name:      DataTypes.STRING(500),
+    email:          DataTypes.STRING(1000),
```
k3ntako: same comment as above about reducing the limit.

server/sequelize/models/Person.js
```diff
@@ -0,0 +1,62 @@
...
+          month: birthdate.getUTCMonth(),
+          year: birthdate.getUTCFullYear(),
+        };
+      },
+      set(val) {
+        this.setDataValue('birthdate', new Date(`${val.year}-${val.month}-${val.date}T00:00:00`))
+      }
+    },
+    sex:            DataTypes.ENUM('M', 'F'),
+    location:       DataTypes.STRING(500),
```
k3ntako: same comment as above about reducing the limit.

server/sequelize/models/PersonSettings.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+module.exports = (sequelize, DataTypes) => {
+  const PersonSettings = sequelize.define('PersonSettings', {
+    send_attendance: DataTypes.BOOLEAN
+  }, {
+    underscored: true,
```
k3ntako: I'm a little confused by you set `underscored` to true.

server/controllers/PersonController/index.js
```diff
@@ -0,0 +1,13 @@
...
+  'listForMyTeam', 
+  'updateMetrics', 
+  'updateSelf', 
+  'updateSettings'
+];
+
+module.exports = METHODS.reduce( (exports, method) => {
+  exports[method] = require(`./${method}`);
+  return exports;
+}, {});
```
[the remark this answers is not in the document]
mienaikoe: If you want to add, remove, or name a method in the future, you just modify it in the array instead of in two places (potential for typo).

server/controllers/PersonController/updateMetrics.js
```diff
@@ -0,0 +1,18 @@
+const { PersonMetric } = require('../../sequelize/models')
+
+module.exports = async function(req, res, next){
+  const values = req.body
+
+  try {
+    await PersonMetric.update(
+      req.body,
```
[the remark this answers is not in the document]
mienaikoe: yea..

server/controllers/PersonController/updateSettings.js
```diff
@@ -0,0 +1,18 @@
+const { PersonSettings } = require('../../sequelize/models')
+
+module.exports = async function(req, res, next){
+  try {
+    await PersonSettings({
```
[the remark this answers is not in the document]
mienaikoe: whoops!

server/sequelize/migrations/20191007223203-create-person.js
mienaikoe: uuids aren't that important unless you need to know for sure if something is a user. I doubt we'll have any instances where this happens (most primary keys I've seen in the wild are just integers.

server/sequelize/migrations/20191007223203-create-person.js
mienaikoe: oh boy naming! lol

so postgres uses varchar, so having something that's designed to be 10x longer than most people will need it isn't a big deal space-wise. It would only be a problem if some hacker started making thousands of accounts, each with the names 500 bytes long. Also unicode might make 500 bytes actually 200-some characters based on which characters are chosen.

I do think we should have a limit on the frontend, but I doubt it will be a problem anytime soon.

server/sequelize/migrations/20191007223203-create-person.js
mienaikoe: I mean it could.. it would just be the kind of input you have to go left and right on. 

Also I wonder if other teams will use this field to put down their full addresses.

But yea. I guess 100 would be more practical.

server/sequelize/models/Person.js
mienaikoe: in postgresql you can't have nested types. This is why I'm introducing separate models for various person-nested models.

Birthdate as on object seemed superfluous, however, because postgresql has an actual Date field that doesn't include time or timezone. so I store it as Date, and parse it to object in the API.

server/sequelize/models/PersonSettings.js
mienaikoe: I want all our fields to be underscore instead of CamelCase because of postgresql's automatic lowercasing issue.

I guess this was a way to both document and enforce that the fields should all be underscored. Hmm maybe I should change the migrations so createdAt and updatedAt are underscore as well.

server/sequelize/models/Person.js
```diff
@@ -0,0 +1,62 @@
...
+
+  Person.inclusions = {};
+  Person.associate = function(models) {
+    Person.hasOne(models.PersonMetric, {
+      foreignKey: 'person_id',
+      as: 'metrics'
+    });
+    Person.hasOne(models.PersonAuth, {
+      foreignKey: 'person_id',
+      as: 'authentication'
```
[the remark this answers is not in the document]
mienaikoe: ooh no, not really. Will rename.

server/sequelize/migrations/20191007223203-create-person.js
mienaikoe: uuids are a go after a convo

PR #146, opened 2019-10-13
Sequelize session

k3ntako:


server/controllers/SessionController/join.js
```diff
@@ -0,0 +1,146 @@
...
+    email = email.trim().toLowerCase();
+
+    const salt = new Date().getTime();
+    const passhash = passhashify(password, salt);
+
+    let person = await Person.
+      findOne({
+        where: { email },
+      }, { 
+        include: { model: Team, as: "teams" },
```
mienaikoe: every guide I've ready has the value of include as an Array. It's also like that in the documentation https://sequelize.org/master/manual/models-usage.html#eager-loading. Does it work as an object? Should we convert to array for consistency with documentation?

server/controllers/SessionController/join.js
```diff
@@ -0,0 +1,146 @@
...
+        include: [{
+          association: User.associations.auth,
+        }, {
+          association: User.associations.metrics,
+        }, {
+          association: User.associations.settings,
+        }],
+      });
+
+      let session = await Session.findOne({ where: { sid: req.session.id }});
```
mienaikoe: Wait. Can't you set the person_id on `req.session` and then the session libraries will persist the session values for you? You shouldn't have to do the find and save manually.

server/controllers/SessionController/join.js
```diff
@@ -0,0 +1,146 @@
...
+    req.session.person_id = person.id;
+    req.session.is_verified = person.is_verified;
+
+    res.json(person.toJSON());
+    next();
+  } catch(err) {
+    console.error(err);
+    next(err);
+  }
+}
```
mienaikoe: This function is gigantic. Can it be chunked out into separate functions that describe their purpose? Especially nice because it has its own file now.

server/controllers/SessionController/join.js
```diff
@@ -0,0 +1,146 @@
...
+    if( !first_name || !last_name ){
+      throw new ValidationError("You must supply a name");
+    }
+    if (!email || !email.trim() || !password ){
+      throw new ValidationError("You must supply an email and password")
+    }
+
+    // all emails in wakebird are case insensitive
+    email = email.trim().toLowerCase();
+
```
mienaikoe: should we verify the email against the regex in EmailUtils?

server/controllers/SessionController/login.js
```diff
@@ -0,0 +1,51 @@
...
+        ],
+      });
+
+    if (!person) {
+      throw new ForbiddenError("Email not found");
+    }
+
+    const auth = person.getAuth();
+    if (!auth || !auth.pass_hash) {
+      throw new ForbiddenError("You haven't joined yet");
```
mienaikoe: I just realized this message isn't the best. We should state that we know they exist, but they have yet to set a password. Hmm I wish we could send a special code so the frontend knows to like render a link to the join page or something. but that's not in scope.

server/controllers/SessionController/logout.js
```diff
@@ -0,0 +1,16 @@
+const { Session } = require('../../sequelize/models');
+
+module.exports = async (req, res, next) => {
+  try {
+    await Session.destroy({ where: { sid: req.session.id } });
+    
+    await new Promise(function (resolve, reject) {
+      return req.session.destroy(resolve);
+    });
```
mienaikoe: Why do you need to destroy it twice?

server/controllers/SessionController/me.js
```diff
@@ -0,0 +1,32 @@
...
+
+module.exports = async (req, res, next) => {
+  try {
+    const person = await Person.findByPk(
+      req.session.person_id,
+      {
+        include: [
+          { model: Team, as: "teams" },
+          { model: PersonMetric, as: "metrics" },
+          { model: PersonSetting, as: "settings" },
```
mienaikoe: PersonMetric and PersonSetting would be undefined here.

server/controllers/SessionController/requestPassReset.js
```diff
@@ -0,0 +1,52 @@
...
+      return SessionUtils.modifySession(session.id, {
+        person_id: null,
+        team_id: null,
+        is_verified: null,
+        is_coach: null
+      });
+    })
+  );
+
+  await Session.destroy({ where: { person_id: person.id } });
```
mienaikoe: This effectively does what the above Promise.all block does. We don't need to clear out a session if we're just going to delete it later.

server/controllers/SessionController/resetPassword.js
```diff
@@ -0,0 +1,54 @@
...
+const { ValidationError, ForbiddenError } = require('../errors');
+
+module.exports = async (req, res, next) => {
+  try {
+    if (req.session.person_id) {
+      throw new ValidationError("You're already logged in");
+    }
+
+    const email = req.body.email && req.body.email.trim().toLowerCase();
+    if (!req.body.email || !req.body.password) {
```
mienaikoe: another good time to validate the email address against a regex. I almost feel like there should be some middleware that validates emails as some type Email. but also out of scope of this ticket.

server/controllers/SessionController/verifyEmail.js
```diff
@@ -0,0 +1,92 @@
...
+    let email = req.body.email && req.body.email.trim().toLowerCase();
+
+    if (!email) {
+      throw new ValidationError("You must supply an email")
+    }
+
+    let person = await Person.
+      findOne(
+        { where: { email } }, 
+        { include: { model: Team, as: "teams" } }
```
mienaikoe: another include that isn't an array

server/controllers/SessionController/join.js
k3ntako: An object seems to work from my test, but the documentation definitely says array so I will change it to an array.

https://sequelize.org/master/class/lib/model.js~Model.html#static-method-findAll
Screen Shot 2019-10-18 at 10 26 09 AM

server/controllers/SessionController/join.js
k3ntako: I opened a new PR for this. Just in case we didn't like how I did it, I didn't wanna deal with rebase and stuff.

server/controllers/SessionController/join.js
```diff
@@ -0,0 +1,149 @@
+const { Person, Session, Team, PersonAuth } = require('../../sequelize/models');
+const EMAIL_REGEX = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
k3ntako: Import regex

mienaikoe: use EmailUtils EMAIL_REGEX

server/controllers/SessionController/join.js
k3ntako: put all function in one file.

server/controllers/SessionController/getSessionTeam.js
```diff
@@ -0,0 +1,18 @@
+const { Team } = require('../../sequelize/models');
+
+module.exports = async (req, res, next) => {  
+  try {
+    let team = null;
+    if (req.session.team_id) {
+      team = await Team.
+        findByPk(req.session.team_id).
+        select(req.session.is_coach ? '' : Team.MEMBER_SELECT). //TODO: use "attributes" instead of "select" when TeamController is completed
+        populate(Team.getPopulator(req.session.is_coach)); //TODO: use "include" instead of "populate" when TeamController is completed
```
k3ntako: Throw error saying TODO, so we don't forget to make changes.

server/controllers/SessionController/join.js
```diff
@@ -0,0 +1,149 @@
...
+
+    const salt = new Date().getTime();
+    const passhash = passhashify(password, salt);
+
+    let person = await Person.
+      findOne({
+        where: { email },
+        include: [{ model: Team, as: "teams" }],
+      }).
+      populate(JOIN_POPULATOR); //TODO: update using `.include` after team controller
```
k3ntako: Throw error saying TODO, so we don't forget to make changes.

server/controllers/SessionController/join.js
```diff
@@ -0,0 +1,149 @@
...
+
+      if( teams.length === 1 ){
+        req.session.team_id = teams[0].id;
+        req.session.is_coach = false; 
+      }
+
+      person = await Person.
+        findByPk( person.id ).
+        select(Person.SELF_SELECT). //TODO: update with `options.attributes` after Person model 
+        populate(DEEP_POPULATOR); //TODO: update using `.include` after team controller
```
k3ntako: Throw error saying TODO, so we don't forget to make changes.

server/controllers/SessionController/setSessionTeam.js
```diff
@@ -0,0 +1,46 @@
+const { Person, Team } = require('../../sequelize/models');
+
+const { ForbiddenError } = require('../../errors');
+
+//TODO: Do this after TeamController is completed
```
k3ntako: Throw error saying TODO, so we don't forget to make changes.

server/controllers/SessionController/join.js
```diff
@@ -0,0 +1,149 @@
+const { Person, Session, Team, PersonAuth } = require('../../sequelize/models');
+const EMAIL_REGEX = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+
+const { generateEmailToken, verifyEmailToken, passhashify } = require('../../utilities/AuthUtils');
+const EmailUtils = require('../../utilities/EmailUtils');
+const { ValidationError } = require('../../errors');
+
+module.exports = async (req, res, next) => {
+  try{
```
mienaikoe: also, use a transaction

server/controllers/SessionController/join.js
```diff
@@ -67,61 +67,63 @@ const createPerson = async (params) => {
...
-
-      await team.save();
-    } else {
-      console.warn("A person with a team is joining that was not invited to that team");
-      /*
-        we will allow this for now, because we don't anticipate it ever happening,
-        but it could be the result of some legacy thing. Additionally, we anticipate
-        removing specific team from the person model eventually.
-      */
-    } // else, somehow this person was added as a member without a password
```
k3ntako: I removed this `console.warn` and it now throws an error. This shouldn't be a problem, right?

server/controllers/SessionController/join.js
```diff
@@ -67,61 +67,63 @@ const createPerson = async (params) => {
 
 const createPersonFromInvite = async (req, person, personAuth, params) => {
```
k3ntako: I only added transactions to `createPersonFromInvite()` and not `createPerson()`, because `createPerson()` only has one transaction (`Person.create()`). Just wanted to double check that that's the correct decision.

server/controllers/SessionController/join.js
mienaikoe: something's fishy about that, Let's talk about it tomorrow.

PR #147, opened 2019-10-13
Bug fix: creating team didn't update 'me'

k3ntako:
`SessionReducer.Methods.setTeam(team)` is used to set current team. However, when creating, the team has to also be added the `me.teams`.


client/pages/TeamForm/TeamFormPage.js
```diff
@@ -23,8 +23,13 @@ function TeamFormPage(props){
 
   function submit(){
     if( isCreating ){
-      Team.create({name, email}).then( team => {
-        props.setTeam(team);
+      Team.create({name, email}).then( newTeam => {
+        let newMe = Object.assign({}, props.me);
```
mienaikoe: new Me( props.me );

client/pages/TeamForm/TeamFormPage.js
```diff
@@ -23,8 +23,13 @@ function TeamFormPage(props){
...
+        if (!newMe.teams.find(team => team.id === newTeam.id) ){
+          newMe.teams.push(newTeam);
+        }
+
+        props.setMe(newMe, newTeam);
```
mienaikoe: Just refresh the value of `me` from the api here.

PR #151, opened 2019-10-30
Sequelize TeamController and SeasonController

mienaikoe:


server/controllers/TeamController/_validator.js
```diff
@@ -0,0 +1,62 @@
...
+    });
+    if( claimedTeam ){
+      throw new ValidationError("Team with this email already exists");
+    }
+
+    return name;
+  },
+  validateEmail: async function(email, teamId){
+    // Validate team email
+    email = email.trim().toLowerCase(); // team email is stored as lowercase
```
k3ntako: should we be using the library to verify this?

server/sequelize/models/SeasonPerson.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+module.exports = (sequelize, DataTypes) => {
+  const SeasonPerson = sequelize.define('SeasonPerson', {
+    role: DataTypes.STRING(20),
```
k3ntako: It looks like you could define this as an emun.  
https://sequelize.org/master/class/lib/data-types.js~ENUM.html

if we define it as an enum, we should be able to access the values through `rawAttributes` according to this post. This is old, and I can't find documentation on `rawAttributes.values`. 
https://stackoverflow.com/questions/22341138/get-sequelize-js-enum-values-from-already-defined-model

I could find only one place that mentioned `rawAttributes`. I can look into this more later.
https://sequelize.org/master/class/lib/sequelize.js~Sequelize.html#static-method-where

If we take this approach you wouldn't have to define `SeasonPerson.ROLES`

server/sequelize/models/SeasonPerson.js
```diff
@@ -0,0 +1,19 @@
...
+    indexes: [
+      { unique: false, fields: [ 'role' ] },
+    ],
+    tableName: 'season_people', // Sequelize defaults to pluralized version of model name (i.e., Sessions)  
+  });
+  SeasonPerson.associate = function(models) {
+    SeasonPerson.belongsToMany(models.Season);
+    SeasonPerson.belongsToMany(models.Person);
+  };
+  SeasonPerson.ROLES = ['COACH', 'MEMBER', 'INVITEE', 'INACTIVE'];
```
k3ntako: See comment above

server/controllers/TeamController/create.js
```diff
@@ -0,0 +1,39 @@
...
+    const email = await validateEmail(req.body.email);
+
+    const team = await sequelize.transaction( async t => {
+      const newOwner = await Person.findByPk(
+        req.session.person_id, 
+        { transaction: t }
+      );
+      if ( !newOwner ){
+        throw new ForbiddenError("Person with the session person ID does not exist");
+      }
```
k3ntako: In addition to this check, should we check if `auth` and `auth.pash_hash` exist?

A little unrelated, but I think we could have a better way to check if they made an account.

server/controllers/TeamController/getMine.js
```diff
@@ -0,0 +1,20 @@
+const { Team, Season } = require('../../sequelize/models');
+
+
+module.exports = async function mine(req, res, next){
+  try{
+    team = await Team.find({
```
k3ntako: I think you mean `findAll`

server/controllers/SeasonController/addPeople.js
```diff
@@ -0,0 +1,133 @@
...
+    where: {
+      email: cleanedEmail
+    },
+    include: [{
+      model: PersonAuth
+    }],
+    defaults: {
+      first_name: "New",
+      last_name: "Paddler"
+    },
```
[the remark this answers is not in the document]
k3ntako: My worry is that someone won't update their name when they create an account and their name ends up as "New Paddler",

server/controllers/SeasonController/removePerson.js
```diff
@@ -0,0 +1,65 @@
...
+
+async function destroySession(person_id){
+  await Session.destroy({
+    where: {
+      person_id
+    },
+    limit: 1,
+    transaction
+  });
+}
```
k3ntako: Instead of `limit: 1`, what if we used `transactions` to undo any destroys with more than one?

We should only be deleting one even without the limit. So, if we end up deleting more than one, there is a problem in our DB, and we probably shouldn't delete anything at that point.

server/controllers/SeasonController/_modifySessions.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+]
+
+
+
+const ROLES = ['coaches', 'members', 'inactives'];
+
+
+
+async function modifySessions(people_ids, modifications){
```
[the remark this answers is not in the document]
mienaikoe: Remove File

server/controllers/SeasonController/addPeople.js
```diff
@@ -0,0 +1,133 @@
...
+            id: person.id
+          }
+        }
+      );
+
+      return `${process.env.APP_URL}/join?vt=${verification_token}&email=${person.email}`;
+    } else {
+      return `${process.env.APP_URL}/join`;
+    }
+  }
```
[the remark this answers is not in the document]
mienaikoe: Inner if statement is unnecessary because verification expiration should never exist if there is no passhash

server/controllers/SeasonController/addPeople.js
```diff
@@ -0,0 +1,133 @@
...
+}
+
+
+async function createMembershipForEmail(season, transaction, email){
+  const cleanedEmail = email.toLowerCase().trim();
```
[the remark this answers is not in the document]
k3ntako: never mind!

server/controllers/TeamController/_validator.js
k3ntako: never mind!

server/controllers/SeasonController/addPeople.js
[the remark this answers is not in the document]
mienaikoe: I'll make this an email object!

server/controllers/SeasonController/addPeople.js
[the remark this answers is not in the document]
mienaikoe: Note 1: Make a readme in SeasonController about how we have invites as regular people even though that's super weird.

Note 2: nullable name fields

server/controllers/SeasonController/removePerson.js
```diff
@@ -0,0 +1,65 @@
...
+    const {person_id} = req.body;
+
+    if( !person_id ){
+      throw new ValidationError("No person_id was specified for removeMember");
+    }
+
+    await sequelize.transaction( async transaction => {
+      await removePersonFromSeason(person_id, season_id, transaction);
+      await destroySession(person_id, transaction);
+    });
```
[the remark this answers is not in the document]
mienaikoe: Message would be about how the frontend should be ashamed and to call us

server/controllers/SeasonController/setRole.js
```diff
@@ -0,0 +1,130 @@
...
+      await Promise.all( 
+        memberships.map( membership => {
+          if( person_ids.includes(membership.person_id) ){
+            if( !membership.person.auth.pass_hash ){
+              throw new ValidationError("You cannot change the role of someone who has not yet joined Wakebird.")
+            }
+            membership.role = role
+            return membership.save({
+              transaction
+            })
```
mienaikoe: Check if there's a way to do this in a query. This is kind of heavy-handed and SQL can definitely support this. Not sure about sequelize but it would be weird if they couldn't.

server/controllers/SeasonController/setRole.js
```diff
@@ -0,0 +1,130 @@
...
+        transaction
+      });
+
+      const coachMembership = memberships.find( membership => membership.person_id === req.session.person_id );
+      if( !coachMembership || coachMembership.role !== 'COACH' ){
+        throw new ValidationError("You must be a coach to add people");
+      }
+
+      const person_ids = req.body.person_ids;
+      if( !person_ids ){
```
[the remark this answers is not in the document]
mienaikoe: ugh python got crossed with js in my head

server/sequelize/models/SeasonPerson.js
mienaikoe: ROLE enums should be individual from migration

server/controllers/TeamController/create.js
mienaikoe: we have these checks in router

server/controllers/SeasonController/removePerson.js
mienaikoe: This is more important for destroying SeasonPerson. Sessions are complicated and we should destroy every session for a person, not just one.

server/controllers/SeasonController/addPeople.js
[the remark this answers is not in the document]
mienaikoe: Took a look at the email address object that is returned from email-addresses and it's pretty rad. ngl. I'll just use that actually.

server/controllers/SeasonController/create.js
```diff
@@ -0,0 +1,41 @@
+const { Season, Team, sequelize } = require('../../sequelize/models');
+const { validateName } = require('./_validator');
+
+
+module.exports = async function(req, res, next){
+  try{
+    const name = validateName(req.body.name);
+    const email = name.toLowerCase().replace("\W+","-");
```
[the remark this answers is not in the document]
mienaikoe: yea I was thinking that's how it would be done. We can use the combo of the team email and the season email to construct a unique email that's also parseable and human-readable!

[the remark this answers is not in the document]
mienaikoe: That said, I think this would be a thing for EmailController to handle once we have both controllers available in the sequelize branch

server/controllers/SeasonController/setRole.js
```diff
@@ -0,0 +1,130 @@
...
+          if( person_ids.includes(membership.person_id) ){
+            if( !membership.person.auth.pass_hash ){
+              throw new ValidationError("You cannot change the role of someone who has not yet joined Wakebird.")
+            }
+            membership.role = role
+            return membership.save({
+              transaction
+            })
+          } else {
+            return null;
```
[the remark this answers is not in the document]
mienaikoe: I think we should error if the number of rows update doesn't match the number of rows supplied. I'll add that

server/controllers/SeasonController/setRole.js
```diff
@@ -0,0 +1,130 @@
...
+      const memberships = await SeasonPerson.findByPk(req.session.season_id, {
+        include: {
+          model: SeasonPerson
+        },
+        transaction
+      });
+
+      const coachMembership = memberships.find( membership => membership.person_id === req.session.person_id );
+      if( !coachMembership || coachMembership.role !== 'COACH' ){
+        throw new ValidationError("You must be a coach to add people");
```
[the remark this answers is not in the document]
mienaikoe: nice catch

server/controllers/SeasonController/removePerson.js
mienaikoe: Also:

Sessions are pretty throwaway. The worst case if you dropped the whole table is that everyone would have to login again. If we delete more or less than 1, then I think that's def cause for alarm, but more of an alarm for us than an alarm for the user. I think I want to remove the limit so that every session is removed, even if there are multiple (like with a phone and a laptop owned by the same person).

server/controllers/SeasonController/setRole.js
```diff
@@ -1,48 +1,57 @@
...
     await sequelize.transaction( async transaction => {
-      const memberships = await SeasonPerson.findByPk(req.session.season_id, {
+      const season = await Season.findByPk(req.session.season_id, {
         include: {
           model: SeasonPerson
         },
         transaction
       });
+      
+      const memberships = season.memberships;
```
k3ntako: I assume `memberships` refers to `SeasonPerson`s. If so, don't we need `as: "membership"` after line 19?

server/sequelize/migrations/20191007223203-create-person.js
```diff
@@ -9,11 +9,9 @@ module.exports = {
         allowNull: false,
       },
       first_name: {
-        allowNull: false,
         type: Sequelize.STRING(300)
```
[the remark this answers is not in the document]
mienaikoe: Most databases save 4 bytes for every primitive and I think postgres is one of them. Also this is a string's length limit, not the limit of values. You're thinking of ints/chars

mienaikoe:
Still need to fix logic for session.is_coach

server/sequelize/models/SeasonPerson.js
```diff
@@ -3,7 +3,7 @@
 module.exports = (sequelize, DataTypes) => {
   const SeasonPerson = sequelize.define('SeasonPerson', {
     role: {
-      type: DataTypes.ENUM(),
+      type: DataTypes.ENUM,
       values: ['COACH', 'MEMBER', 'INVITEE', 'INACTIVE'],
```
k3ntako: Doesn't `values` go inside the `()`? Like this:
```
  DataTypes.ENUM({
    values: ['COACH', 'MEMBER', 'INVITEE', 'INACTIVE']
  })
```
https://sequelize.org/v5/class/lib/data-types.js~ENUM.html

mienaikoe: I think it's different for models for some reason https://sequelize.org/master/manual/data-types.html

mienaikoe: fixed

PR #152, opened 2019-11-03
Sequelize LocationController

k3ntako:


k3ntako:
Should locations belong to a team or a season? I think we said team, but I'm wondering if it would be better if it's on the season. If someone puts their home address on there and leaves the team, I don't think they want that info to be around 10 years later.

server/sequelize/models/location.js
```diff
@@ -0,0 +1,15 @@
...
+  const Location = sequelize.define('Location', {
+    name: DataTypes.STRING,
+    latitude: DataTypes.DOUBLE,
+    longitude: DataTypes.DOUBLE,
+    address: DataTypes.STRING,
+    note: DataTypes.JSONB,
+    google_id: DataTypes.STRING,
+  }, {});
+  Location.associate = function(models) {
+    Location.belongsTo(models.Team);
```
mienaikoe: Location should belong to a Person

if we want to add to team, that'd be a whole feature and we could add that in later.

server/sequelize/migrations/20191026000527-create-location.js
```diff
@@ -0,0 +1,49 @@
...
+      },
+      address: {
+        type: Sequelize.STRING
+      },
+      note: {
+        type: Sequelize.JSONB
+      },
+      google_id: {
+        type: Sequelize.JSONB
+      },
```
mienaikoe: remove this field

server/sequelize/migrations/20191020165421-create-season-people.js
```diff
@@ -35,7 +35,7 @@ module.exports = {
         return Promise.all([
           queryInterface.addConstraint(
             TABLE_NAME, ['person_id', 'season_id'], {
-              type: 'primary key',
+              type: 'PRIMARY KEY',
```
k3ntako: revert this!

mienaikoe: https://gist.github.com/lucasscariot/5b8747fbc8a6948a805c646fae4ceef8

server/sequelize/migrations/20191026000527-create-location.js
k3ntako: I'm looking at the front-end code and it looks like you were using the `google_id` for Map stuff. Our production DB is storing this info. 

### /client/components/EventMap/index.js
```
 onSingleClick(ev){
    this._recentClick = null;
    const placeid = ev.placeid;
    if( placeid ){
      // clicking on the marker isn't important to catch
      if( placeid === this.props.location.google_id ){
        return;
      }

      // clicked on a place on the map
      this.geocoder.geocode({
        bounds: this.state.bounds,
        placeId: placeid
      }, (results) => {
        this.onPlaceChanged(results[0]);
      });

    ...
  }
```

mienaikoe: oh snap! :| good catch

I feel like using the google_id to detect a market click isn't the best, but maybe a problemf or the future.

server/sequelize/migrations/20191020165421-create-season-people.js
```diff
@@ -31,11 +31,11 @@ module.exports = {
         }
       }
     ).then( () => {
-      return queryInterface.sequelize.transaction(function (t) { 
+      return queryInterface.sequelize.transaction(function (t) {
         return Promise.all([
           queryInterface.addConstraint(
             TABLE_NAME, ['person_id', 'season_id'], {
-              type: 'PRIMARY KEY',
+              type: 'primary key',
```
mienaikoe: if you check the link I posted, I think there's still an easier way to do this by marking `primary_key: true` when defining the fields instead of afterward with `addConstraint`

mienaikoe:
lgtm

PR #155, opened 2019-11-16
Sequelize event

k3ntako:


server/controllers/EventController/find.js
```diff
@@ -0,0 +1,59 @@
...
+      order: [
+        [{ model: EventPeople, as: 'attending' }, 'first_name', 'ASC'],
+        [{ model: EventPeople, as: 'attending' }, 'last_name', 'ASC'],
+        [{ model: EventPeople, as: 'not_attending' }, 'first_name', 'ASC'],
+        [{ model: EventPeople, as: 'not_attending' }, 'last_name', 'ASC'],
+      ],
+    });
+
+    if (!event) {
+      throw new NotFoundError();
```
k3ntako: I think we were saying NotFound should be just for pages. I noticed that we were using it in our old code, and after looking into it I think 404 is appropriate.

404 is indicates that the:
- "browser was able to communicate with a given server, but the server could not find what was requested." ([Wikipedia](https://en.wikipedia.org/wiki/HTTP_404))
- "server can't find the requested resource." ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404))

I feel like rows in a database counts as "what was requested" or a "resource".

server/controllers/EventController/attend.js
```diff
@@ -0,0 +1,43 @@
...
+          is_attending: req.body.is_attending,
+          event_id: req.params.event_id,
+          person_id: req.session.person_id,
+        }, {
+          transaction,
+        });
+
+        if (!eventPerson) throw new ValidationError("Failed to create event person");
+      } else if (updateCount[0] > 1) {
+        throw new ValidationError(`Only one EventPerson should be updated. ${updatedCount} event people were updated and rolled back.`);
```
k3ntako: I have a lot of these checks. These should never happen unless something is wrong with our database.
1. Should this be a 500 error instead of the 400s? Client did nothing wrong.
2. It's nice that we are catching this, but I would be so mad if I got this error as a user. The user is completely powerless. Making the same request probably will result in the same issue, and the developers (us) aren't informed about this so we can't really fix these. Should we have something to alert us when this happens (maybe in a PR next year)?

server/controllers/EventController/create.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    // Create assignments
+    const assignments = (req.body.assignments || []).map(async assignment => {
+      const name = assignment.name.trim();
+      const assignee_limit = assignee_limit || 1;
+      if (!name) throw new ValidationError('Invalid assignment name');
+      if (!Number.isInteger(assignee_limit) || assignee_limit < 1){
+        throw new ValidationError('Assignee limit must be a positive integer');
+      }
```
k3ntako: Instead of throwing an error, should I just convert it to `1`?

k3ntako:
This should be merged after SeasonController, because migrations mention `seasons`.

server/controllers/EventController/find.js
mienaikoe: Hmm ok. I guess "page" is a bit outdated.

server/controllers/EventController/attend.js
mienaikoe: Yea this seems like a 500. I think at the very least, a user might report this error to us.

server/controllers/EventController/create.js
mienaikoe: nah. The UI should not allow for this to happen. If it does come in as a string, or come in as like negative numbers. I think we would want to know, as engineers so we can fix the problem quicker.

There definitely is a problem of user errors vs engineer errors that we probably aren't addressing well, but we can get to that later. I think there's a simple change we could do on the frontend to handle this.

server/controllers/EventController/attend.js
k3ntako: **Note to self**

line 34: `${updatedCount}` should be `${updatedCount[0]}`.

server/controllers/EventController/delete.js
```diff
@@ -0,0 +1,33 @@
...
+    await sequelize.transaction(async transaction => {
+      const destroyCount = await Event.destroy({
+        where: {
+          id: req.params.event_id,
+        },
+        transaction,
+      });
+
+      if (destroyCount[0] === 0) {
+        if (!eventPerson) throw new ValidationError("Event was not deleted");
```
k3ntako: **Note to self**

Delete `if (!eventPerson)`

server/controllers/EventController/delete.js
```diff
@@ -0,0 +1,33 @@
...
+      } else if (destroyCount[0] > 1) {
+        throw new ValidationError(`Only one Event should be deleted. ${updatedCount} events were deleted and rolled back.`);
+      }
```
k3ntako: I think I'm just bikeshedding, but I wanted to get your opinion on something if you don't mind.

Line 13:
```
if (destroyCount[0] === 0) {
```
Destroy count can be 0 for two reasons. 
1. Client gave an invalid `req.params.event_id`.
2. Server had an issue.

The first one is a 400,  but the second one is a 500. Should we be differentiating between those like the following? Or is that too much work?
```
const event = await Event.findAll(...);

if(!event.length) throw new ValidationError(...);

const destroyCount = await event[0].destroy();
if(destroyCount[0] === 0) throw new ServerError(...);
```

server/controllers/EventController/delete.js
```diff
@@ -0,0 +1,34 @@
...
+        throw new ServerError(`Only one Event should be deleted. ${destroyCount[0]} events were deleted and rolled back.`);
+      }
+
+      // No need to check destroyCount, because destroying 0 and destroying more than 1 are both valid
+      await EventPerson.destroy({
+        where: {
+          event_id: req.params.event_id,
+        },
+        transaction,
+      });
```
k3ntako: maybe this should a before or after hook?

server/controllers/EventController/delete.js
mienaikoe: A database orm like sequelize shouldn't give 0 rows back if it encounters an issue. It should throw an exception. And I think you can rely on it to do so. I think it's fine the way it's written now.

server/controllers/EventController/delete.js
mienaikoe: yea I'm all about hooks. This looks like a `beforeDestroy`. You'll get an error if you destroy a row has foreign keys pointed at it.

server/controllers/EventController/_validator.js
```diff
@@ -0,0 +1,19 @@
+const parseBody = ({ name, datetime, location_id, note }) => ({
+  name: name && name.trim(),
+  datetime: (datetime || datetime === 0) && new Date(datetime), // new Date(null).getTime() === new Date(0).getTime() is true
+  location_id: location_id && location_id.trim(),
+  note,
+});
```
mienaikoe: beautiful request handling!

server/controllers/EventController/attend.js
```diff
@@ -0,0 +1,43 @@
...
+
+      if (updateCount[0] === 0) {
+        // First time RSVPing for this event
+        const eventPerson = await EventPerson.create({
+          is_attending: req.body.is_attending,
+          event_id: req.params.event_id,
+          person_id: req.session.person_id,
+        }, {
+          transaction,
+        });
```
mienaikoe: const [record, created] = await Model.upsert(
  { id: 1, name: 'foo' }, // Record to upsert
  { returning: true }     // Return upserted record
);

https://stackoverflow.com/questions/29063232/sequelize-upsert

server/controllers/EventController/attend.js
```diff
@@ -0,0 +1,43 @@
...
+
+        if (!eventPerson) throw new ServerError("Failed to create event person");
+      } else if (updateCount[0] > 1) {
+        throw new ServerError(`Only one EventPerson should be updated. ${updateCount[0]} event people were updated and rolled back.`);
+      }
```
mienaikoe: no need to check when your where clause uses a primary key

server/controllers/EventController/attendanceOverride.js
```diff
@@ -0,0 +1,37 @@
...
+      }
+
+      const updateCount = EventPerson.update({
+        is_attending: req.body.is_attending,
+      }, {
+        where: {
+          event_id: req.params.event_id,
+          person_id: req.body.people,
+        },
+        transaction,
```
mienaikoe: specify a transaction

server/controllers/EventController/attendanceOverride.js
```diff
@@ -0,0 +1,37 @@
...
+      });
+
+      if (updateCount[0] !== req.body.person.length){
```
mienaikoe: peoples'

server/controllers/EventController/attendanceOverride.js
```diff
@@ -0,0 +1,37 @@
...
+        console.warn(`Number of people updated, ${updateCount[0]}, does not equal the number of people specified ${req.body.person.length}`);
```
mienaikoe: throw this. 

Also, peoples'

server/controllers/EventController/cancellation.js
```diff
@@ -0,0 +1,34 @@
...
+          season_id: req.session.season_id,
+        },
+        transaction
+      });
+
+      if (updateCount[0] === 0) {
+        throw new ServerError('Event was not updated');
+      } else if (updateCount[0] > 1) {
+        throw new ServerError(`Only one Event should be updated. ${updateCount[0]} events were updated and rolled back.`);
+      }
```
mienaikoe: another extra check when using pk

server/controllers/EventController/create.js
```diff
@@ -0,0 +1,56 @@
...
+      season_id: req.session.season_id,
+    }, {
+      include: [{
+        model: Assignment,
+        as: 'assignments',
+        include: [{
+          model: Person,
+          as: 'assignees',
+        }]
+      }],
```
mienaikoe: no one is assigned anything at creation time

server/controllers/EventController/delete.js
```diff
@@ -0,0 +1,34 @@
...
+        where: {
+          id: req.params.event_id,
+        },
+        transaction,
+      });
+
+      if (destroyCount[0] === 0) {
+        throw new ServerError("Event was not deleted");
+      } else if (destroyCount[0] > 1) {
+        throw new ServerError(`Only one Event should be deleted. ${destroyCount[0]} events were deleted and rolled back.`);
```
mienaikoe: extra with pk

server/controllers/EventController/delete.js
mienaikoe: Look into Cascade

server/controllers/EventController/find.js
```diff
@@ -0,0 +1,65 @@
...
+          is_attending: false,
+        },
+      }, {
+        model: Boat,
+        as: 'boats',
+        where: boatWhere,
+        include: [{
+          model: Person,
+          as: 'seats',
+        }],
```
mienaikoe: You're gonna want a BoatPerson (Many to Many relationship)

server/controllers/EventController/index.js
```diff
@@ -0,0 +1,15 @@
+const METHODS = [
+  'create',
+  'find',
+  'delete',
+  'cancellation',
+  'listForMyTeam',
```
mienaikoe: listForMySeason

server/controllers/EventController/update.js
```diff
@@ -0,0 +1,79 @@
...
+  );
+}
+
+
+module.exports = async (req, res, next) => {
+  try {
+    if (!req.params.event_id) {
+      throw new ValidationError('Invalid Event ID');
+    }
+
```
mienaikoe: do this all in a transaction

server/controllers/EventController/update.js
```diff
@@ -0,0 +1,79 @@
...
+    await Promise.all(
+      req.body.assignments.map(async assignmentJson => {
+        if (assignmentJson.id) {
+          return await updateAssignemnt(assignmentJson);
```
mienaikoe: typo

server/controllers/EventController/update.js
```diff
@@ -0,0 +1,79 @@
...
+            }]
+          }
+        ],
+        transaction,
+      });
+
+      if (updateCount[0] === 0) {
+        throw new ServerError("Event was not updated");
+      } else if (updateCount[0] > 1) {
+        throw new ServerError(`Only one Event should be updated. ${updateCount[0]} events were updated and rolled back.`);
```
mienaikoe: extra

server/sequelize/migrations/20191115150527-create-event-people.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('event_people', {
+      id: {
+        type: Sequelize.DataTypes.UUID,
+        defaultValue: Sequelize.literal('uuid_generate_v4()'),
+        primaryKey: true,
+        allowNull: false,
+      },
```
mienaikoe: you can leave this off for a compound id

server/sequelize/migrations/20191115150527-create-event-people.js
```diff
@@ -0,0 +1,43 @@
...
+          key: 'id',
+        },
+      },
+      person_id: {
+        type: Sequelize.UUID,
+        references: {
+          model: 'person',
+          key: 'id',
+        },
+      },
```
mienaikoe: Compound Primary Key

server/sequelize/models/Event.js
```diff
@@ -0,0 +1,31 @@
...
+
+  Event.associate = function(models) {
+    Event.belongsTo(models.Season);
+    Event.belongsTo(models.Location);
+    Event.hasMany(models.Person, {
+      foreignKey: 'event_id',
+      through: {
+        model: models.EventPerson,
+      },
+    });
```
mienaikoe: need hasMany accessor for EventPerson as well

server/sequelize/models/Person.js
```diff
@@ -52,20 +52,26 @@ module.exports = (sequelize, DataTypes) => {
...
       foreignKey: 'owner_id',
       as: 'teams_owned'
-    })
+    });
+    Person.hasMany(models.Event, {
+      foreignKey: 'person_id',
+      through: {
+        model: models.EventPerson,
+      },
+    });
```
mienaikoe: box it to EventPerson (for the season recap grid)

server/controllers/EventController/find.js
```diff
@@ -0,0 +1,70 @@
...
+      }, {
+        model: Boat,
+        as: 'boats',
+        where: boatWhere,
+        include: [{
+          model: BoatPerson,
+          as: 'seats',
+          include: [{
+            model: BoatPerson,
+            as: 'seats',
```
k3ntako: model should be Person and as should be person

server/sequelize/models/Event.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+module.exports = (sequelize, DataTypes) => {
+  const Event = sequelize.define('Event', {
+    name: DataTypes.STRING(255),
```
k3ntako: missing ID

server/sequelize/models/EventPerson.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+module.exports = (sequelize, DataTypes) => {
+  const EventPerson = sequelize.define('EventPerson', {
+    is_attending: DataTypes.BOOLEAN,
```
k3ntako: add event_id and person_id

server/sequelize/migrations/20191104191129-create-event.js
```diff
@@ -0,0 +1,70 @@
...
+      },
+      note: {
+        type: Sequelize.JSONB,
+      },
+      season_id: {
+        type: Sequelize.UUID,
+        references: {
+          model: 'season',
+          key: 'id',
+        },
```
k3ntako: cascade delete

k3ntako: `allowNull: false`

server/controllers/EventController/_validator.js
```diff
@@ -0,0 +1,19 @@
...
+
+
+module.exports = (body) => {
+  const { name, datetime, location_id, note } =  parseBody(body);
+
+  if (!name) throw new ValidationError('Invalid event name');
+  if (!datetime || Number.isFinite(datetime.getTime())) {
```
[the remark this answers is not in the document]
mienaikoe: ugh why did they name that function `isFinite`. 

If you're just checking for NaN, you should use `Number.isNaN`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

server/controllers/EventController/_validator.js
```diff
@@ -0,0 +1,19 @@
+const parseBody = ({ name, datetime, location_id, note }) => ({
+  name: name && name.trim(),
+  datetime: (datetime || datetime === 0) && new Date(datetime), // new Date(null).getTime() === new Date(0).getTime() is true
```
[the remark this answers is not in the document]
mienaikoe: I think you can simplify this:

```
function isValidDate(d) {
  return d instanceof Date && !isNaN(d);
}
```

https://stackoverflow.com/questions/1353684/detecting-an-invalid-date-date-instance-in-javascript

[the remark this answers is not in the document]
k3ntako: Don't allow dates before 1970. check if number is greater than 0.

server/controllers/EventController/update.js
k3ntako: `updateAssignemnt` should be `updateAssignment`

server/controllers/EventController/_validator.js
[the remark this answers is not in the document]
k3ntako: I'm thinking we can let dates less than 0 (aka less than 1970). Since it's a util that might be used in other places, I think it should be less opinionated.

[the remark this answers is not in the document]
mienaikoe: ok. As long as you also supply the bounds check elsewhere that needs it

server/controllers/EventController/_validator.js
```diff
@@ -0,0 +1,26 @@
...
+  const { name, datetime, location_id, note } =  parseBody(body);
+
+  if (!name) throw new ValidationError('Invalid event name');
+  if (!datetime) throw new ValidationError('Invalid event datetime');
+  if (!location_id) throw new ValidationError('Invalid location ID');
+
+  const midnightThisMorning = new Date().setHours(0, 0, 0, 0);
+  if (datetime.getTime() < midnightThisMorning) {
+    throw new ValidationError('Start date for this event cannot be in the past');
+  }
```
k3ntako: This will throw an error for any `datetime`s that older than the midnight of that day. This will be used by `create` and `update`

Should we allow `update` to change the date to something in the past? I'm thinking no, because Wakebird is not a diary but  a way to keep people up-to-date on future events.

PR #157, opened 2019-11-29
Sequelize BoatController

k3ntako:


server/controllers/BoatController/clone.js
```diff
@@ -0,0 +1,38 @@
...
+      event_id: req.params.event_id,
+    });
+
+    const boatPersonPromises = templateBoat.seats.map(seat  => {
+      return BoatPerson.create({
+        name: seat.name,
+        boat_id: seat.boat_id,
+        person_id: seat.person_id,
+      });
+    });
```
mienaikoe: Use a transaction for this operation. As a general rule, you want to use a transaction for anything that has more than one create or update. Also a generally decent rule to use for more than one retrieval, but deadlock is less likely in that context.

server/controllers/BoatController/find.js
```diff
@@ -0,0 +1,26 @@
...
+      throw new ValidationError("No boat ID passed in")
+    }
+
+    const boat = await Boat.findByPk(req.params.boat_id, {
+      include: [{
+        model: BoatPerson,
+        as: 'seats',
+        include: [{
+          model: Person,
+          attributes: ["id", "first_name", "last_name"],
```
mienaikoe: Coach would also want a person's set of metrics (weight, erg times) for the balance chart.

server/controllers/BoatController/seat.js
```diff
@@ -0,0 +1,40 @@
+const { Sequelize, BoatPerson } = require('../../sequelize/models');
+const { ValidationError } = require('../errors');
+
+module.exports = async (req, res, next) => {
+  try {
+    if (!req.params.boat_id && !req.body.person_id) {
+      throw new ValidationError("Expected boat ID and/or person ID");
```
mienaikoe: I think boat_id is required. person_id is optional. In this case you don't need the `&&`

server/controllers/BoatController/seat.js
```diff
@@ -0,0 +1,40 @@
...
+      }
+    });
+
+    if (assignSeat) {
+      await BoatPerson.create({
+        boat_id: req.params.boat_id,
+        person_id: req.body.person_id,
+        name: req.params.seat_name,
+      })
+    }
```
mienaikoe: Suggest using a transaction

server/sequelize/migrations/20191129013949-create-boat-person.js
```diff
@@ -0,0 +1,53 @@
...
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('boat_people', {
+      id: {
+        type: Sequelize.DataTypes.UUID,
+        defaultValue: Sequelize.literal('uuid_generate_v4()'),
+        primaryKey: true,
+        allowNull: false,
+      },
+      name: {
```
mienaikoe: what do you think about renaming this field to "`seat_name`"?

server/sequelize/migrations/20191129013949-create-boat-person.js
```diff
@@ -0,0 +1,53 @@
...
+        allowNull: false,
+        type: Sequelize.ENUM(SEATS),
+      },
+      boat_id: {
+        allowNull: false,
+        type: Sequelize.UUID,
+        references: {
+          model: 'event',
```
mienaikoe: should be model: 'boat'

server/sequelize/migrations/20191129013949-create-boat-person.js
```diff
@@ -0,0 +1,53 @@
...
+          key: 'id',
+        },
+        onDelete: 'cascade',
+      },
+      person_id: {
+        allowNull: false,
+        type: Sequelize.UUID,
+        references: {
+          model: 'event',
```
mienaikoe: should be model: 'person'

server/sequelize/models/BoatPerson.js
```diff
@@ -0,0 +1,38 @@
...
+  '3', '3L', '3R',
+  '4', '4L', '4R',
+  '5', '5L', '5R',
+  '6', '6L', '6R',
+  '7', '7L', '7R',
+  '8', '8L', '8R',
+  '9', '9L', '9R',
+  '10', '10L', '10R',
+  'S',
+];
```
mienaikoe: is it possible to keep this enum in one place so both the migration and model can reference it?

server/sequelize/models/BoatPerson.js
```diff
@@ -0,0 +1,38 @@
...
+      type: Sequelize.ENUM(SEATS),
+    },
+    boat_id: {
+      allowNull: false,
+      type: DataTypes.UUID,
+    },
+    person_id: {
+      allowNull: false,
+      type: DataTypes.UUID,
+    },
```
mienaikoe: compound primary key. add `primaryKey: true` to both of these.

server/sequelize/models/BoatPerson.js
k3ntako: I think the issue is that the model can change but migration can't. So, if we ever change the enum, the original migration that uses the enum will break.

k3ntako: BoatPerson can have a pre-create hook that checks for the enum instead of having it in the migration and model.

Migration and model are rigid and we expect to change this if we add other types of boats.

server/sequelize/migrations/20191129013949-create-boat-person.js
k3ntako: instead of `as: seats`, change it to `as: boatPeople`

server/controllers/BoatController/_getBoatPersonInclude.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  return {
+    model: BoatPerson,
+    as: 'boatPeople',
+    include: [{
+      model: Person,
+      attributes: ["id", "first_name", "last_name"],
+    }],
+  }
```
mienaikoe: nit: something I've learned recently from Flatiron is to put large data structures into variables, and then separate them from the logic that selects them. An example of this would be:
```
const COACH_INCLUDE = ...
const DEFUALT_INCLUDE = ...
module.exports = function(isCoach){
  return isCoach ? COACH_INCLUDE : DEFAULT_INCLUDE;
} 
```

This makes your code a lot more readable

server/controllers/BoatController/clone.js
```diff
@@ -0,0 +1,47 @@
...
+        }],
+      });
+
+      boat = await Boat.create({
+        name: `Copy of ${templateBoat.name}`,
+        event_id: req.params.event_id,
+      }, {
+        include: [
+          getBoatPersonInclude(req.session.is_coach),
+        ]
```
mienaikoe: you need to supply the transaction variable to each of these queries. Transactions don't come for free.

server/controllers/BoatController/listForEventId.js
```diff
@@ -0,0 +1,25 @@
...
+
+module.exports = async (req, res, next) => {
+  try {
+    if (!req.params.event_id) {
+      throw new ValidationError("No event ID specified")
+    }
+
+    const boats = await Boat.findAll({
+      where: {
+        event_id: req.params.event_id
```
mienaikoe: might want to include: `hidden: (isCoach ? null : false)`

server/sequelize/models/BoatPerson.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  BoatPerson.beforeBulkCreate((instances, options) => {
+    instances.forEach(instance => seatNameChecker(instance, options));
+  });
+  BoatPerson.beforeBulkUpdate((options) => seatNameChecker(options.attributes, options));
+  BoatPerson.beforeCreate(seatNameChecker);
+  BoatPerson.beforeUpdate(seatNameChecker);
+  BoatPerson.beforeSave(seatNameChecker);
+  BoatPerson.beforeUpsert((values, options) => seatNameChecker(values, options, true));
```
mienaikoe: Do you need this to be applied for all of these? I figured beforeSave would cover all of these cases, but I didn't look it up.

k3ntako: I changed it to `beforeValidate`, seems to be called every time.

I couldn't find anything definitive, but I think `beforeSave` is only called for `.save()`.
In their [documentation](https://sequelize.org/v5/manual/hooks.html), the order of operation has it as a peer in step 4.

PR #159, opened 2019-12-14
Fix Controller and Model Syntax that gets in the way of Tests

mienaikoe:


server/tests/test.js
```diff
@@ -18,14 +18,14 @@ test({
 */
 
 
-module.exports = async function test({testMethod, req, assertions, done}){
+module.exports = async function test({method, req, assertions, done}){
```
k3ntako: in the example you should change `testMethod` to `method`

PR #160, opened 2019-12-14
PersonController Tests

mienaikoe:
The tests work!!

Had to make a few changes to models to get it there. Learned a lot in the process. Let's talk about it later.


server/tests/test.js
```diff
@@ -32,8 +32,8 @@ module.exports = async function test({method, req, assertions, done}){
       done("Method responded with something falsey. Controller methods must respond with JSON Objects.");
     } else if(!responseRaw.constructor) {
       done(`Method responded with a primitive. Controller methods in Wakebird must respond with JSON Objects.`)
-    } else if(responseRaw.constructor.name !== 'Object'){
-      done("Method responded with a non-Object. Controller methods must respond with JSON Objects");
+    } else if(responseRaw.constructor.name !== 'Object' && responseRaw.constructor.name !== 'Array'){
```
k3ntako: For array, I think we can use `Array.isArray()`

server/tests/TestHelper.js
```diff
@@ -21,7 +21,7 @@ async function join(email){
     let person = await Person.create({
       first_name: "Test",
       last_name: "User",
-      email: email || `TestUser${randomnum()}@wakeBird.com`,
+      email: email || `testuser${randomnum()}@wakebird.com`,
```
k3ntako: I always assumed that we had these weird capitalization to make sure that the our server would lower case them.

server/controllers/PersonController/listForMySeason.js
```diff
@@ -0,0 +1,21 @@
...
+module.exports = async function(req, res, next){
+  try {
+    const people = await Person.findAll({
+      include: [{
+        model: Season,
+        required: true,
+        where: {
+          id: req.session.season_id
+        },
+      }]
```
k3ntako: This `include` would include the season on every person. I thought there was a way to query this without the include, but I can't seem to find it.

server/controllers/PersonController/updateMetrics.js
```diff
@@ -1,17 +1,16 @@
...
   try {
     const updateFields = mapUpdateFields(req.body)
-    await PersonMetric.update(
+    updateFields.person_id = req.params.person_id;
+    await PersonMetric.upsert(
       updateFields,
-      {
-        where: { person_id: req.params.person_id }
-      }
+      { returning: false }
```
k3ntako: Without the where, how will `Sequelize` know what to update? Does it update the rows with . the matching `person_id`?

What if none of the `person_id`s match, but the `erg200m` matches, wouldn't it update that?

server/controllers/PersonController/updateSettings.js
```diff
@@ -1,13 +1,11 @@
...
-    await PersonSettings.update({
-      send_attendance: req.body.attendance_notifications
-    },{
-      where: {
-        person_id: req.session.person_id
-      }, 
+    await PersonSettings.upsert({
+      send_attendance: req.body.send_attendance,
+      person_id: req.session.person_id
     });
```
k3ntako: Same as above

server/tests/PersonController/find.js
```diff
@@ -0,0 +1,40 @@
...
+
+
+module.exports = function(){
+  let test_session;
+
+  before(async function(){
+    test_session = await TestHelper.join(test_email);
+  })
+  
+  it('updates metrics for a person', function(done) {
```
k3ntako: i think the description should be something different. this is for `PersonController.find`

server/tests/PersonController/listForMySeason.js
```diff
@@ -0,0 +1,35 @@
...
+
+
+module.exports = function(){
+  let test_session;
+
+  before(async function(){
+    test_session = await TestHelper.join();
+  })
+  
+  it('updates metrics for a person', function(done) {
```
k3ntako: same as above

server/tests/PersonController/listForMySeason.js
```diff
@@ -0,0 +1,35 @@
...
+      assertions: async function(response){
+        expect(response.length).to.be.greaterThan(0);
+        response.forEach( person => {
+          assert(person.id);
+          assert(person.email);
+          expect(person.auth).to.be.undefined;
+          expect(person.metrics).to.be.undefined;
+          expect(person.settings).to.be.undefined;
+          expect(person.seasons).to.be.undefined;
+        });
```
k3ntako: should we be more specific? We could test the number of people on the team, or the names of the people?

server/tests/PersonController/updateSelf.js
```diff
@@ -0,0 +1,55 @@
...
+
+
+module.exports = function(){
+  let test_session;
+
+  before(async function(){
+    test_session = await TestHelper.join();
+  })
+  
+  it('updates a person', function(done) {
```
k3ntako: maybe it could be more descriptive: "updates a person's own information"

server/tests/PersonController/updateMetrics.js
```diff
@@ -0,0 +1,48 @@
...
+      method: PersonController.updateMetrics, 
+      req:{
+        session: test_session,
+        body: metricBody,
+        params: {
+          person_id: test_session.person_id
+        }
+      }, 
+      assertions: async function(response){
+        assert(response.okay)
```
k3ntako: I believe `assert` just checks for truthy-ness. Instead we can check if it is specifically `true` using [`assert.isTrue`](https://www.chaijs.com/api/assert/#method_istrue).

server/tests/PersonController/updateSelf.js
```diff
@@ -0,0 +1,55 @@
...
+    test({
+      method: PersonController.updateSelf, 
+      req:{
+        session: test_session,
+        body: updateBody,
+      }, 
+      assertions: async function(response){
+        assert(response.okay)
```
k3ntako: same as above, `assert.isTrue()`

server/tests/PersonController/updateSettings.js
```diff
@@ -0,0 +1,39 @@
...
+  
+  it("updates a person's own settings", function(done) {
+    test({
+      method: PersonController.updateSettings,
+      req:{
+        session: test_session,
+        body: updateBody,
+      }, 
+      assertions: async function(response){
+        assert(response.okay)
```
k3ntako: same as above, `assert.isTrue()`

server/controllers/PersonController/updateMetrics.js
mienaikoe: have since refactored out metrics.

server/controllers/PersonController/updateSettings.js
mienaikoe: Sequelize uses unique keys to decipher an update vs insert. Will add a unique index if not already existant.

mienaikoe: verified unique key already exists on this table

server/tests/PersonController/updateMetrics.js
mienaikoe: refactored metrics, and no other methods so far return okay: true

server/tests/PersonController/updateSettings.js
mienaikoe: updateSettings now returns the settings model because it has the potential to upsert

server/sequelize/migrations/20191215232856-create-metric-type.js
```diff
@@ -0,0 +1,50 @@
...
+      },
+      unit: {
+        type: Sequelize.ENUM(
+          'POUNDS', 'KILOGRAMS',            // Personal Weight, Weightlifting
+          'SECONDS',                        // Erg or Run (Should be displayed as MM:SS)
+          'METERS', 'WATTS',                // Erg
+          'KILOMETERS', 'MILES',            // Running, Swimming, Biking
+          'FEET', 'CENTIMETERS',            // Personal Height or Wingspan
+          'LITERS', 'CALORIES', 'GRAMS',    // Nutrition
+          'PERCENT', 'REPS',                // Reps or Goals
```
k3ntako: I noticed that you don't have milliseconds. Just to clarify, we can just use seconds if we need milliseconds (e.g., 100 meter dash), because it looks like the metric is a float so.

k3ntako: NIT: I think we could have inches instead of feet. It's probably easier going from inches to feet than vice versa.

k3ntako: We were initially saying that we only need one unit for each type of measurement. For example, we can use meters to get kilometers. Couldn't we do that? We can just assume that they want the largest units possible? For a run it might be `3621m`, and we can just show that as `3.62 km`. I think this is especially true for SI units where conversion between them are very easy. 

We should keep both SI and customary units so that we don't have to save their preferences.

k3ntako:
Initially, we avoided capital letters in our database, but since we are using them now for references, should we just use them for table names too? It's definitely what Sequelize wants us to do.

server/sequelize/models/Team.js
```diff
@@ -18,6 +18,8 @@ module.exports = (sequelize, DataTypes) => {
     Team.hasMany(models.Season, {
       foreignKey: 'team_id',
       as: 'seasons',
```
k3ntako: can we remove this `as`?

k3ntako:
This is slightly unrelated, but I noticed that in your PersonController files (`find.js`, `listForMySeason.js`, `updateSettings.js`), you don't have an `await` before `toJSON()`. 

I'm a little confused because the documentation seems to suggest that `toJSON` returns a promise, but your tests are passing. https://sequelize.org/v5/class/lib/model.js~Model.html#instance-method-toJSON

Is there a reason why you didn't add an `await` before `toJSON`?

server/sequelize/migrations/20191215232856-create-metric-type.js
mienaikoe: I think some measurements should be specified because it allows the coach to specify whatever they want. If the coach wants to measure things in feet, then we can give them feet and inches inputs. If the coach wants to measure things only in inches (maybe it's like the length of their paddle), then they can specify that. We shouldn't be in the business of dictating what unit should be used for what metric.

I'm thinking similar if people specify they want to measure things in seconds or minutes or hours. You give them their major unit, and their minor unit (example: `05h 32m`, `7m 34s`, `55.9 sec`, `3ft 5in"` `55.00 inches`, etc). I think the minor unit can be saved in the decimal because we can dictate which figure to round to for each unit.

Because of this, I think we should permit as many units as we think are likely to be used, regardless of whether they can be inferred from another unit. I'll add both inches and feet. I really don't think milliseconds would ever be used as a major unit tho.

server/sequelize/models/Team.js
mienaikoe: Yea. I want to avoid `as` as much as possible because Sequelize dictates that we use it in both the model and the query to get them joined with some other model.

I am very unhappy with how opinionated sequelize is about the capitalization of our table names, and how difficult they made it to customize the table names. Postgres will change case on you, which is the original reason I wanted to make everything lowercase. But this shouldn't come at the cost of putting `as` in all of our queries because this is likely to cause bugs where we forget to include it.

Therefore, my thinking is as follows:
- Do not use `as`. Instead, accept the CamelCasing of all foreign key references.
- Use lowercase for all other fields (typically primitives like varchar, int, and boolean)

This has the added benefit of being able to easily differentiate between model fields and primitive fields.

server/sequelize/models/MetricType.js
```diff
@@ -0,0 +1,29 @@
...
+    description: DataTypes.STRING,
+    unit: DataTypes.ENUM({
+      values: [
+        'POUNDS', 'KILOGRAMS',            // Personal Weight, Weightlifting
+        'SECONDS',                        // Erg or Run (Should be displayed as MM:SS)
+        'METERS', 'WATTS',                // Erg
+        'KILOMETERS', 'MILES',            // Running, Swimming, Biking
+        'FEET', 'CENTIMETERS',            // Personal Height or Wingspan
+        'LITERS', 'CALORIES', 'GRAMS',    // Nutrition
+        'PERCENT', 'REPS',                // Reps or Goals
```
k3ntako: This doesn't seem up-to-date with the migration

server/sequelize/models/Person.js
```diff
@@ -36,34 +40,15 @@ module.exports = (sequelize, DataTypes) => {
...
-
-    Person.hasMany(models.EventPerson, {
-      foreignKey: 'person_id',
-    });
-
+    Person.hasMany(models.SeasonPerson);
+    Person.hasMany(models.Team);
+    Person.hasMany(models.EventPerson);
     Person.belongsToMany(models.Event, {
       foreignKey: 'person_id',
```
k3ntako: NIT: should we remove this, just in case we forget in the future? or is that out of the scope for this PR?

k3ntako:
NIT: I wonder if `MeasurementType` would be more appropriate than `MetricType`. Metric type makes me think that we are defining units. 

This tells us **what we are measuring** and not **what metric we are using**.

k3ntako:
Maybe I missed it, but it doesn't look like you have API endpoints for creating/editing/deleting metric types. Is that something you wanna do in this PR, or in the future?

server/sequelize/migrations/20191215232856-create-metric-type.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+
+const ALL_UNITS = [
+  // Personal Weight, Weightlifting
+  'POUNDS',     // 155 lb
+  'KILOGRAMS',  // 85 kg
+
+   // Erg or Run Times (Should be displayed as MM:SS)
+  'HOURS', // 2h 5m
```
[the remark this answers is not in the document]
mienaikoe: Yes.

server/tests/PersonController/find.js
```diff
@@ -0,0 +1,40 @@
...
+        body: null,
+        params: {
+          person_id: test_session.person_id
+        }
+      }, 
+      assertions: async function(response){
+        assert(response.id);
+        assert.equal(response.email, test_email);
+        assert.equal(response.first_name, 'Test');
+        assert.equal(response.last_name, 'User');
```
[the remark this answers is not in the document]
mienaikoe: Why?

server/sequelize/models/MetricType.js
mienaikoe: I'll refactor to keep the enum in one place.

server/sequelize/models/Person.js
mienaikoe: Why would we remove this?

mienaikoe: Why would we remove this?

server/tests/PersonController/find.js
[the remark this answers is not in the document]
k3ntako: Considering that expect and assert generally have alternatives for each other, we can use one to accomplish what we are doing. I think it would be easier to read if we have one consistent way of writing tests

server/sequelize/models/MetricType.js
k3ntako: wouldn't that cause issues if we ever change the enum?

server/sequelize/models/Person.js
k3ntako: Weren't we removing `foreignKey`s and letting Sequelize figure it out?

server/sequelize/models/MetricType.js
mienaikoe: you can't solve both problems. You either have to:
1) Accept that the migration and type will be unmatched sometimes
2) Accept that the migration will be updated, but not necessarily run for update reasons.
3) Store the field as a varchar in the migration, and then make it an enum in the model.

server/sequelize/models/Person.js
mienaikoe: ah ok. I think only `as` is important to remove since it changes the behavior of the association. `foreignKey` doesn't, but isn't really necessary so will remove.

mienaikoe: hmm these are already removed in my version

server/tests/PersonController/find.js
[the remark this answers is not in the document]
mienaikoe: I guess I disagree. I'm encountering this at work rn. One of our engineers has a different philosophy of defining private variables than we do. We came to the conclusion that it really doesn't matter because we can all read both approaches and telling people to change the format of their tests and relearn another because of nothing more than visual perfection is a waste of time.

server/sequelize/models/MetricType.js
k3ntako: I'm worried that this could cause some problems in the future. 

Maybe this is overkill, but I wonder if we could have dated migration files for the enum. The file will just export the enum, we would make a new file every time we change the enum.

If you think that the approach above is overkill and you think that changing the enum in migration is safe enough, I trust your judgement!

server/sequelize/models/Person.js
k3ntako: that's strange. maybe you didn't push the latest commits? let me know if there are any updates!

server/tests/PersonController/find.js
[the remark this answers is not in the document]
k3ntako: yeah that makes sense. it definitely was a nitpick.

server/sequelize/models/Team.js
k3ntako: Yeah I definitely agree about Sequelizing being opinionated and how it's annoying

server/sequelize/models/Person.js
```diff
@@ -38,9 +38,9 @@ module.exports = (sequelize, DataTypes) => {
 
   Person.inclusions = {};
   Person.associate = function(models) {
-    Person.hasMany(models.PersonMetric);
+    Person.hasMany(models.PersonMeasurement);
     Person.hasOne(models.PersonAuth);
-    Person.hasOne(models.PersonSettings);
+    Person.hasMany(models.PersonSettings);
```
k3ntako: in what situation would we have more than one `PersonSetting`?

server/tests/PersonController/index.js
```diff
@@ -5,7 +5,24 @@ const ENDPOINTS = [
...
+
+  after(async function(){
+    await Models.PersonSettings.destroy({where: {}});
+    await Models.PersonMeasurement.destroy({where: {}})
+    await Models.SeasonPerson.destroy({where: {}});
+    await Models.Team.update({
+      active_season_id: null
+    },{
+      where: {}
+    });
```
k3ntako: Looks like you are no longer have `active_season_id` on Team

server/controllers/SeasonController/getActive.js
```diff
@@ -0,0 +1,24 @@
...
+      {
+        include: [{
+          model: SeasonPerson,
+          include: {
+            model: Person
+          }
+        }, {
+          model: Team
+        }]
+      });
```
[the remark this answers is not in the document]
mienaikoe: in my mind, a `find` is a method for finding something by its primary key. This is finding the currently active season by using the session's built in season_id parameter.

There can be multiple active seasons, but there is only one season that a user can "use" at a time. This is handled by giving the user a list of seasons and having them set one, which then persist to the session. This would be done through an endpoint that may or may not be created yet in `SessionController`. There's also behavior to automatically set their session to the one active season if the person only has one active season.

server/controllers/SessionController/_populateSession.js
```diff
@@ -1,30 +1,27 @@
-const { sequelize, Sequelize, Season } = require('../../sequelize/models');
+const { sequelize, Sequelize, Season, SeasonPerson } = require('../../sequelize/models');
 
 
 module.exports = async function(session, person){
   session.person_id = person.id;
   session.is_verified = person.is_verified;
 
-  const seasons = await Season.findAll({
+  const season_people = await SeasonPerson.findAll({
```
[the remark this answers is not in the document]
mienaikoe: I've been preferring snake case for local variables. Mostly because it's a habit from python programming, which prefers snake case.

server/controllers/SessionController/_populateSession.js
```diff
@@ -1,30 +1,27 @@
...
+        is_active: true
+      },
+      required: true
     }
   });
 
-  if (seasons.length === 1) {
-    const season = seasons[0];
-    session.team_id = seasion.team.id;
+  if (season_people.length === 1) {
```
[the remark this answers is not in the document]
mienaikoe: Technically, yes. The database doesn't guarantee it.

Effectively, no. The code we have would prevent such a case. But there's always the possibility that a bug sneaks in.

server/sequelize/models/MetricType.js
mienaikoe: I think the field should be a varchar in the migration. This solves both of our concerns.

server/sequelize/models/Person.js
mienaikoe: Ah ok. Ironed it out. Think I was reading things wrong.

So for some reason, the normal Camel Case fields don't work in some cases. I couldn't figure out why, and I far prefer to specify the exact spelling of the fields in question so the field to use is "documented" somewhere. I'm going to keep them in.

server/sequelize/migrations/20191020163316-add-team-fks.js
```diff
@@ -11,15 +11,6 @@ module.exports = {
...
-    const refSeasonDefinition = {
-      type: Sequelize.UUID,
-      allowNull: true,
-      references: {
-        model: 'seasons',
-        key: 'id'
-      }
-    };
-
     return queryInterface.sequelize.transaction(function (t) {
```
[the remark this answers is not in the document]
mienaikoe: yea they were vestigial from a prior iteration

server/sequelize/models/Person.js
mienaikoe: I briefly changed PersonSetting to be a key-value store and then changed my mind. I forgot to set this back to its original state.

server/tests/PersonController/index.js
mienaikoe: yes. I was running into a bunch of issues trying to find active season by joining the team and then reverse-referencing the season. I think we don't really need a single active season to be specified precisely in the database. We can programmatically control that. Plus it gives us the option to change our minds in the future, and simplifies queries immensely by not having to do that reverse reference to get an active season.

server/sequelize/models/MetricType.js
k3ntako: Yeah I think that'd work. We are doing something like that for boat seat names. We have a validation in a hook (only certain seat names are allowed), so it acts like an enum.

server/controllers/SeasonController/getActive.js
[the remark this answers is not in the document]
k3ntako: I was misunderstanding what this endpoint was doing. Sorry, that's my fault.

server/controllers/SessionController/_populateSession.js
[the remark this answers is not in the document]
k3ntako: ahh okay

server/controllers/SessionController/_populateSession.js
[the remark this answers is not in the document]
k3ntako: Do you think a check is necessary here in case of bugs?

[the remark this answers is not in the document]
mienaikoe: I'm not sure what we would do if it was zero. I don't think it's appropriate to throw an error because this method is used by other controller methods that have their own behaviors. I think having no seasons would be a state that the frontend should handle. If there's no season set, the frontend will check for all settable seasons. If many are found, it should display them for selection. If none are found, it should display a message saying they need to be added to a season.

server/sequelize/models/MetricType.js
k3ntako: @mienaikoe No rush, but I thought you might have missed this comment thread. I noticed that you're still using `MeasurementTypeEnum`.

server/tests/TestHelper.js
```diff
@@ -31,38 +31,30 @@ async function join(email){
...
   let join_session = {};
-  join_session.destroy = cb => {
+  join_session.destroy = async function(){
+    await person.removeSeason(getGlobalSeason());
+    await person.destroy();
     delete join_session.person_id
     delete join_session.team_id
     delete join_session.season_id
     delete join_session.is_verified
     delete join_session.destroy
```
k3ntako: My understanding is that we did this because we maintained one session when we tested Mongo. If so, is it unnecessary now to delete everything from the session object?

PR #161, opened 2019-12-15
Creating New PersonMetricController and Tests

mienaikoe:


server/tests/PersonMetricController/listForMetricType.js
```diff
@@ -0,0 +1,55 @@
...
+    test({
+      method: PersonMetricController.listForMetricType, 
+      req:{
+        session: test_session,
+        params: {
+          metric_type_id: test_metric_type.id
+        }
+      },
+      assertions: async function(response){
+        assert(Array.isArray(response));
```
k3ntako: I think the error message for this will be something like "Expected truthy value, got something falsey," which doesn't tell us much. You could add a message as the second argument in `assert`:

`assert(Array.isArray(response), "Response should be an array");`

server/tests/PersonMetricController/listForMetricType.js
```diff
@@ -0,0 +1,55 @@
...
+    })
+  })
+
+  after(async function(){
+    await test_metric_type.destroy();
+    // should cascade to person metric too
+  })
+  
+  
+  it('lists metrics for me', function(done) {
```
k3ntako: I think you meant `'lists metrics for metricType'`

server/tests/PersonMetricController/listForPersonSeason.js
```diff
@@ -0,0 +1,64 @@
...
+    })
+  })
+
+  after(async function(){
+    await test_metric_type.destroy();
+    await test_other_person.destroy();
+    // should cascade to person metric too
+  })
+  
+  it('lists metrics for me', function(done) {
```
k3ntako: something like `'lists metrics for a person's season'`?

mienaikoe:
I like expect because it's more versatile and reads like plain english.

server/controllers/PersonMetricController/listForMetricType.js
```diff
@@ -0,0 +1,21 @@
...
+        required: true,
+      }]
+    });
+
+    res.json(metrics.map(metric => metric.toJSON()));
+    next();
+  } catch(err) {
+    next(err)
+  }
+}
```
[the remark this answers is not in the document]
mienaikoe: Would be used by coach.

This would not reveal other teams' info, because a metric type is tied to a Team's Season.

server/tests/PersonMetricController/listForMetricType.js
```diff
@@ -0,0 +1,55 @@
...
+      unit: "KILOGRAMS",
+      SeasonId: test_session.season_id,
+    })
+    test_person_metric = await PersonMetric.create({
+      person_id: test_session.person_id,
+      value: 10.0,
+      note: 'Nothing to Note',
+      measurement_date: "Sun Dec 15 2019 14:24:12 GMT-0500 (Eastern Standard Time)",
+      MetricTypeId: test_metric_type.id
+    })
```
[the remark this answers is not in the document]
mienaikoe: 2 wouldn't give anything extra other than to check that we are searching for more than one, which imo is a weak test.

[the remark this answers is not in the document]
mienaikoe: realized that this would be testing whether it's filtering by person, which is important, so it was added.

server/tests/PersonMetricController/_global.js
```diff
@@ -0,0 +1,58 @@
...
+  await test_other_person.destroy();
+})
+
+module.exports = {
+  globalSession: () => test_session,
+  globalMetricType: () => test_metric_type,
+  globalPersonMetric: () => test_person_metric,
+  globalOtherPerson: () => test_other_person,
+  globalOtherPersonMetric: () => test_other_person_metric,
+};
```
k3ntako: NIT: I think you mean this only for the PersonMetricController, I initially thought global meant we were using these for all controller tests. This isn't a big deal, so feel free to resolve this.

server/tests/PersonMetricController/index.js
```diff
@@ -1,9 +1,10 @@
 const ENDPOINTS = [
+  "_global",
```
k3ntako: did you mean to add `_global` here? It's not an endpoint, so it doesn't belong here?

server/tests/PersonMetricController/_global.js
mienaikoe: hmm maybe global isn't the best word to use here. I also didn't want it to just random say `Session()`

Maybe `getMockSession()` would be more appropriate

server/tests/PersonMetricController/index.js
mienaikoe: true. Don't need to export it.

server/tests/PersonMetricController/_global.js
mienaikoe: Moving to "fixtures" language

server/controllers/SeasonController/index.js
```diff
@@ -1,10 +1,7 @@
 const SeasonController = {
-  'addPeople': require('./addPeople'),
   'create': require('./create'),
   'getActive': require('./getActive'),
-  'getMine': require('./getMine'),
-  'removePerson': require('./removePerson'),
-  'setRole': require('./setRole'),
+  'setActive': require('./setActive'),
```
mienaikoe: To be called by a coach to set the currently active season when the season flips over.

server/controllers/SeasonPersonController/index.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  'create': require('./create'),
+  'getMine': require('./getMine'),
+  'delete': require('./delete'),
+  'update': require('./update'),
+};
```
mienaikoe: The changes in this SeasonPersonController don't really belong here, but I thought it wise to do this while I remember to do it. The SeasonController was getting quite long and I felt it needed to be broken out into things that deal with the season specifically, and things that deal with the SeasonPerson models.

server/tests/MeasurementTypeController/listForMySeason.js
```diff
@@ -0,0 +1,30 @@
...
+    const test_session = fixtures.getSession();
+    const test_measurement_type = fixtures.getMeasurementType();
+
+    test({
+      method: MeasurementTypeController.listForMySeason, 
+      req:{
+        session: test_session,
+      }, 
+      assertions: async function(response){
+        expect(response.length).to.equal(1); // global coach and this new person
```
k3ntako: looks like this comment is a copy and paste

server/controllers/MeasurementTypeController/listForMySeason.js
```diff
@@ -0,0 +1,19 @@
+const {  
+  MeasurementType
+} = require('../../sequelize/models');
+
+
+module.exports = async function(req, res, next){
+  try {
+    const measurement_types = await MeasurementType.findAll({
+      where: {
+        season_id: req.session.season_id
```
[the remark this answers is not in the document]
mienaikoe: No. See talk about foreign key in some other comment.

server/tests/MeasurementTypeController/listForMySeason.js
```diff
@@ -0,0 +1,30 @@
...
+      method: MeasurementTypeController.listForMySeason, 
+      req:{
+        session: test_session,
+      }, 
+      assertions: async function(response){
+        expect(response.length).to.equal(1);
+        response.forEach( measurement_type => {
+          expect(measurement_type.name).to.eql(test_measurement_type.name)
+          expect(measurement_type.unit).to.eql(test_measurement_type.unit);
+          expect(measurement_type.description).to.eql(test_measurement_type.description);
```
k3ntako: NIT: It probably doesn't matter, but the documentation for [`eql`](https://www.chaijs.com/api/bdd/#method_eql) says: "Asserts that the target is deeply equal to the given obj." It basically does a deep equal check instead of `===`.

[`equal`](https://www.chaijs.com/api/bdd/#method_equal) seems to be `===`. It's not a big deal, and the only reason I bring it up is because for a second I thought they were objects when I saw `eql`.

mienaikoe: yeh i've been using `eql` everywhere instead of `equal`, since deeply equal is more powerful than `===`. The only case I can imagine needing `equal` instead of `eql` is like when you want to evaluate whether an object or array is the exact same reference as something else (which is should be very rare for the kinds of tests we are writing)

server/controllers/SeasonController/setActive.js
```diff
@@ -0,0 +1,34 @@
...
+        req.params.season_id,
+        { transaction }
+      )
+
+      if( !targetSeason.is_active ){
+        await Season.update(
+          { is_active: false },
+          { 
+            where: {
+              TeamId: targetSeason.TeamId
```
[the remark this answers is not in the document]
mienaikoe: see convo about foreign keys in other comment (I forget where)

PR #164, opened 2019-12-25
LocationController tests for Postgres

k3ntako:
The tests are failing because our `server/tests/test.js` function doesn't allow arrays. I noticed that you addressed in your PR, so I didn't change it here.

Another issue with that function is that it doesn't allow us to test for errors. For example, if the user is unauthorized, the controller should throw an error. I added a field called `expectToThrow`. I wasn't sure if would be better to have a separate method that we would use when expecting the controller to throw an error.


mienaikoe:
If you want to use functionality on another PR, you can rebase your PR off of it

server/tests/LocationControllerTests/create.js
```diff
@@ -0,0 +1,73 @@
...
+    test({
+      method: LocationController.create,
+      req: {
+        session: test_session,
+        body: locationParams,
+      },
+      assertions: async (response) => {
+        // find location in database
+        const location = await Location.findByPk(response.id);
+        const locationJSON = location.toJSON();
```
mienaikoe: You don't need to test raw JSON when testing that something was persisted to the database.

server/tests/LocationControllerTests/create.js
```diff
@@ -0,0 +1,73 @@
...
+    test_session = await TestHelper.join(email);
+    locationParams = TestHelper.newLocationParams(test_session);
+  });
+
+  after(async () => {
+    // destroy all locations
+    Location.destroy({
+      where: {}
+    });
+  });
```
mienaikoe: You could do this as an `afterEach` at the `index.js` level under `LocationController`

server/tests/test.js
```diff
@@ -18,30 +18,40 @@ test({
...
       responseRaw = responseRawJson;
     }
   }, async function(err){
-    if( err ){
+    if (expectToThrow && !err) {
+      done("Expected method to throw an error.");
+    } else if (expectToThrow && responseRaw) {
+      done("Expected response to be null.");
+    } else if (err && expectToThrow) {
+      await testAssertions(null, err, done);
```
mienaikoe: `testAssertions` don't seem 100% necessary if you're expecting your method to throw. I would only call testAssertions if they exist

server/tests/LocationControllerTests/update.js
```diff
@@ -0,0 +1,139 @@
...
+          name: newName,
+          note: newNote,
+        },
+        params: {
+          location_id: originalLocation.id
+        }
+      },
+      expectToThrow: true,
+      assertions: async (response, error) => {
+        expect(error.message).to.equal('Location ID did not match any of your locations');
```
mienaikoe: I don't think error messages should be tested. You can test the Error class if you want to verify that system errors aren't being bubbled up, but we really don't care about that at such an early juncture.

server/controllers/LocationController/create.js
```diff
@@ -30,7 +30,7 @@ module.exports = async function (req, res, next) {
       longitude: longitude,
       note: note,
       google_id: google_id,
-      person_id: req.session.person_id,
+      PersonId: req.session.person_id,
```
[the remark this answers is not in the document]
k3ntako: I thought we said it would be owned by a person, but maybe I'm misremembering. 

My concern is that if a user doesn't own it, who would have the permission to edit and delete the location? Also, I think this could be problematic if a user wants to delete their home address or something like that.

But, as you mention it would be nice to have a list of locations for a team. It would be overkill to have locations belong to a user AND have a `SeasonLocation`s table, right?

server/tests/TestHelper.js
```diff
@@ -68,6 +69,58 @@ async function join(email){
...
+    PersonId: test_session.person_id,
+  }
+};
+
+async function createLocation(test_session, customParams = {}){
+  const locationParams = Object.assign({}, newLocationParams(test_session), customParams);
+  const location = await Location.create(locationParams);
+
+  return await location.toJSON();
+}
```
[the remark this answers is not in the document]
k3ntako: So, I was thinking we would need this for an event too, because an event has a location. How do you think I should deal with that?

server/tests/LocationControllerTests/delete.js
```diff
@@ -0,0 +1,79 @@
...
+        }
+      },
+      assertions: async (response, error) => {
+        expect(error.message).to.equal('Location was not destroyed');
+
+        const location = await Location.findOne({
+          where: {
+            id: locationFailToDelete.id,
+          }
+        });
```
k3ntako: should use `findByPk`

server/tests/LocationControllerTests/create.js
```diff
@@ -0,0 +1,73 @@
...
+
+  it('should return a location', function(done) {
```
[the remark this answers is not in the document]
k3ntako: do you think i should just combine the two tests?

server/tests/test.js
k3ntako: I don't think I quite understand what you mean here. 

Even if the method throws, we might there might be assertions to run. For example, for `EventController.delete`, we would want to check the database to see if the event still exists. Does that make sense, or maybe that's not what you were talking about.

server/controllers/LocationController/create.js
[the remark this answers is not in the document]
mienaikoe: yea I think that'd be too much. I think you can have an owner and an "owning season" that is nullable. Private locations have no owning season, and public ones do.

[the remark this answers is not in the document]
mienaikoe: Also that's not necessarily in scope for this as it's not something we've validated as a real need.

server/tests/LocationControllerTests/create.js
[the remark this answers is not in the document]
mienaikoe: it's fine either way. I see benefits and drawbacks to both.

server/tests/TestHelper.js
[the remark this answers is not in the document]
mienaikoe: true.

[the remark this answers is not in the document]
mienaikoe: Also, just noticed. You should *not* return the JSON representation, because you will want the caller to explicitly destroy the model after its done using it. Return the original instead.

server/tests/test.js
mienaikoe: key word there is "might". We shouldn't assume that testAssertions is truthy. It may be unspecified, so you should wrap the call in an `if( testAssertions )`

k3ntako: I think you mean that `assertions` might not be specified. `testAssertions` is defined in the file, so it should always be defined. 

Was `testAssertions confusing. I meant it as "test" the verb, but it sounds like the noun. Maybe `runAssertions` is clearer?

server/tests/TestHelper.js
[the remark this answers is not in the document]
k3ntako: I moved the TestHelper functions for Location to a separate file. Let me know what you think.

server/tests/test.js
mienaikoe: yea that's what I mean. method name is fine. Just am easily confused these days.

server/controllers/LocationController/update.js
```diff
@@ -19,14 +23,14 @@ module.exports = async function (req, res, next) {
       }, {
         where: {
           id: req.params.location_id,
-          person_id: req.session.person_id,
+          PersonId: req.session.person_id,
```
mienaikoe: hrm... This sucks, but could you make this lowercase again? I'm sorry. I know it's annoying. I just think the pattern of keeping primitives camel case is a good one.

server/tests/LocationControllerTests/create.js
```diff
@@ -0,0 +1,66 @@
...
+      req: {
+        session: test_session,
+        body: locationParams,
+      },
+      assertions: async (response) => {
+        // find location in database
+        const dbLocation = await Location.findByPk(response.id);
+
+        expect(response).to.have.all.keys([
+          'id', 'name', 'address', 'longitude', 'latitude', 'note', 'google_id', 'person_id', 'PersonId', 'createdAt', 'updatedAt'
```
mienaikoe: it should have either person_id or PersonId, not both.

server/tests/TestHelper/LocationHelper.js
```diff
@@ -0,0 +1,41 @@
+const {
```
mienaikoe: Out-Of-Scope:

Maybe in another ticket we can put the collection of helpers into their own directory. They seem to be cumbersome to leave in one file. We used to do that because we would use models between tests. But I think that's a faux pas in this new regime where our tests are mostly standalone (except for the globals which are just super helpful to have around)

server/tests/TestHelper.js
```diff
@@ -59,6 +60,56 @@ async function join(email){
...
+    note: blankDraftJSNote(),
+    google_id: "ChIJM0-VXKz3wokRZMNlFCQJmEc",
+    PersonId: test_session.person_id,
+  }
+};
+
+async function createLocation(test_session, customParams = {}){
+  const locationParams = Object.assign({}, newLocationParams(test_session), customParams);
+  return await Location.create(locationParams);
+}
```
mienaikoe: you've moved all this code to the other file, but it's still here.

server/tests/LocationControllerTests/update.js
k3ntako: I initially agreed with you that we shouldn't test Error messages, but now I'm thinking we should. A controller could throw an error for the wrong reason and we won't catch it. You could argue the end result is the same, an error. But, the front-end will get an unhelpful message like "location is undefined" instead of "The specified location ID did not match any locations".

We would be testing the error message, not because the specific wording is important, but because we want to make sure that the error being thrown is the one we expected. 

What do you think?

mienaikoe: then you should test for the error class. not the error message.

server/tests/test.js
```diff
@@ -1,32 +1,60 @@
...
+    done();
+  } catch (err) {
+    done(err);
+  }
+}
+
+module.exports = async function test({ method, req, expectToThrow, assertions, done}){
+  if (!method) return done("Controller method not specified");
+  if (!req) return done("Request not specified");
+  if (!done) return done("Done not specified");
```
mienaikoe: this won't work. if done is falsey, you can't call it like a function. You should throw instead.

server/tests/TestHelper/LocationHelper.js
k3ntako: I tried moving each function to its own file. let me know what you think.

server/tests/LocationControllerTests/create.js
k3ntako: So, I think there is no way for us to prevent `PersonId` to be returned. 

This one talks about the issue where both are being returned. In their example, `integration_id` and `integrationId` are both present.
https://github.com/sequelize/sequelize/issues/10857

It seems like `underscored` has many related issues. They consolidated them to one Issue (linked below), but I don't think we can expect an update in the near future.
https://github.com/sequelize/sequelize/issues/11225

I considered specifying the attributes, but that only works on `find` and not `create`. I don't think it's worth making that change if we can't be consistent.
https://stackoverflow.com/questions/54929244/how-to-return-only-specific-attributes-when-using-sequelize-create-method

server/tests/TestHelper/index.js
```diff
@@ -60,6 +60,23 @@ async function join(email){
   return join_session;
 }
```
mienaikoe: Move to a separate helper called `SessionHelper`.

Rename this directory to `helpers`

so in the end you should have:

tests/
  helpers/
     SessionHelper.js
     LocationHelper.js

PR #166, opened 2020-01-03
Done should be called with Error instance

k3ntako:
Maybe didn't have to be a separate PR, but I think I need to work on making my PRs smaller.


k3ntako:
I added a way to have `before` and `after` an `it` block. 

Out `it` blocks are not `async`, so we have to wrap `test` in a `Promise` in order to manipulate the database prior to one test. 

We also do not have a way to run a method after a test, aside from putting it in `assertions`. I think putting things in `assertions` that don't belong isn't idea.

Let me know what you think.

mienaikoe:
before and after should be done with mocha, not in a custom method here. If you need a special before and after, use a describe block, as it describes a different setup case.

server/tests/test.js
```diff
@@ -21,25 +21,30 @@ test({
...
+  const safeDone = (err) => {
+    const error = err instanceof Error ? err : new Error(err);
+    done(error);
+  }
+
+  if (!method) return safeDone("Controller method not specified");
+  if (!req) return safeDone("Request not specified");
   // assertions has to be defined if test is NOT expected to throw
-  if (!expectToThrow && !assertions) return done("Assertions not specified");
+  if (!expectToThrow && !assertions) return safeDone("Assertions not specified");
```
k3ntako: Wanted to get your opinion. I'm on the fence on this,  but I think `assertions` should always exist because even if we expect it to error, we should check the type of the error.

mienaikoe: sounds good

server/tests/test.js
```diff
@@ -49,21 +54,21 @@ module.exports = async function test({ method, req, expectToThrow, assertions, d
     }
   }, async function(err){
     if (expectToThrow && !err) {
-      done("Expected method to throw an error.");
+      safeDone("Expected method to throw an error.");
     } else if (expectToThrow && responseRaw) {
-      done("Expected response to be null.");
+      safeDone("Expected response to be null.");
     } else if (err && expectToThrow) {
```
mienaikoe: You should nest these in one big 
`if (expectToThrow)`

mienaikoe:
> 
> 
> I added a way to have `before` and `after` an `it` block.
> 
> Out `it` blocks are not `async`, so we have to wrap `test` in a `Promise` in order to manipulate the database prior to one test.
> 
> We also do not have a way to run a method after a test, aside from putting it in `assertions`. I think putting things in `assertions` that don't belong isn't idea.
> 
> Let me know what you think.

I don't quite understand what you mean. `before` and `after` were made for these cases. Let's talk about it today.

server/tests/test.js
```diff
@@ -48,22 +55,24 @@ module.exports = async function test({ method, req, expectToThrow, assertions, d
...
-    } else if (err && expectToThrow) {
-      await testAssertions(null, err, assertions, done); // run assertions if expected to throw
-    } else if (err)  {
-      done(err);
+    if (expectToThrow) {
+      if (!err) {
+        safeDone("Expected method to throw an error.");
+      } else if (responseRaw) {
+        safeDone("Expected response to be null.");
+      } else if (err) {
```
mienaikoe: this can just be `else` because you already checked if err wasn't null

PR #167, opened 2020-01-09
EventController Tests

k3ntako:
**There is another [PR](https://github.com/mighty-speed-krush/Mighty-Speed-Krush/pull/168) that might be worth looking at first.** That you don't have to read the parts that will be overwritten.

I cannot test `listForSeries`, because `SeriesController`/`Series` model has not been completed. However, this is not an issue as this API endpoint is not being used. It's not 100% clear what we would use this for as well.


mienaikoe:
You'll have to guide me through this one. It's too long to do alone.

server/sequelize/models/Event.js
```diff
@@ -34,6 +34,15 @@ module.exports = (sequelize, DataTypes) => {
...
+      through: {
+        model: models.EventPerson,
+      },
+    });
+
+    Event.belongsToMany(models.Person, {
+      as: 'NonAttendees',
       foreignKey: 'event_id',
       through: {
         model: models.EventPerson,
```
mienaikoe: I don't think the distinction between attending and not attending is important to make in the backend. You can let the frontend sort them out after receiving. We did these lists in mongo because that's the best way to organize the model in a nosql setting.

server/controllers/EventController/update.js
```diff
@@ -27,55 +30,81 @@ module.exports = async (req, res, next) => {
...
+        model: Location,
+        required: true,
+        attributes: ['id', 'name', 'latitude', 'longitude', 'note', 'address'],
+      }, {
+        model: Assignment,
+        attributes: ["id", "name", "assignee_limit"],
+        include: [{
+          model: Person,
+          attributes: ["id", "first_name", "last_name"],
+          through: { attributes: [] },  // removes AssignmentPerson join table
```
mienaikoe: I think this is dangerous. It breaks the expectation of Assignment's relationships.

server/tests/TestHelper/EventHelper/createEventPerson.js
```diff
@@ -0,0 +1,9 @@
+const { EventPerson } = require('../../../sequelize/models');
+
+module.exports = async function (testSession, eventId, isAttending = true) {
+  return await await EventPerson.create({
```
mienaikoe: double await?

server/controllers/EventController/update.js
k3ntako: I think I'm not sure why this is dangerous. This query just gets the event to be returned to the front-end, and I don't see why we would need the `AssignmentPerson` join table to be sent back.

Maybe the comment was confusing? When I say "removes" I meant removes from the JSON, not the database.

mienaikoe: Yes from the JSON. I think the models we send back through JSON should be predictable every time and should map back to real entities in the database. Even at the expense of sending a superfluous join table. This is a practice that's generally good for APIs because it doesn't require the user of an API to know the intricacies of each endpoint. Even if we're the ones using the API, we should keep our API design clean because our future selves will inevitably forget what our present selves are doing.

Also, we might later want to add a field to AssignmentPerson. Like a note or something. In order to give that to the frontend, we would have to re-introduce this model in between Assignment and Person which would change how our frontend code works quite a bit.

server/controllers/EventController/index.js
```diff
@@ -1,12 +1,13 @@
 const METHODS = [
+  'attend',
+  'attendanceOverride',
+  'cancellation',
```
mienaikoe: These could be moved into an EventPersonController if you're looking for a way to break this into two PRs

k3ntako: I already did all of it here, so I won't break it up unless you think I should

mienaikoe: ugh. Yea time-wise just keep it here I guess. Just look out for these possibilities as you move forward. Breaking things into smaller bits allows us to fit more smaller PRs into our busy schedules.

mienaikoe: tbf, I don't think this would take that long to do after this PR is merged. You move the controller methods, move the tests, rewrite the index files, and change the controller in routes.js. No real thought behind it. Just busy work.

k3ntako: I think I'm not fully understanding what you're saying. You think it would be better to have a separate controller, right? Is the reason because it would be easier to read and/or is it because it would have made the PR smaller? Or something else?

mienaikoe: all of the above

PR #168, opened 2020-01-09
Move inclusions out

k3ntako:
I wasn't sure that this is a good approach so I didn't put it into the test PR.

The reason I didn't put it in the model is because it would make the model massive. Also, no two Event API endpoint has the same inclusion, but they have parts that overlap. So, having multiple files was necessary.

I thought this approach keeps the controller files shorter, and allows us to keep one consistent inclusion for the EventControllers. For example, if we choose that longitude and latitude are not necessary, we can change fewer files.


server/controllers/EventController/_inclusion/findInclude.js
```diff
@@ -0,0 +1,48 @@
...
+  through: {
+    attributes: [], // removes EventPerson join table
+    where: { is_attending: true }, // only attendees
+  },
+  required: false,
+  order: [
+    [{ model: Person }, 'first_name', 'ASC'],
+    [{ model: Person }, 'last_name', 'ASC'],
+  ],
+}
```
mienaikoe: seems weird to see these outside of the main function, because they give details about the query that shouldn't be separated from the rest of the query.

mienaikoe: I think you're overdoing it here. In addition to spreading the query into multiple places, this doesn't add any extra readability that a function couldn't do. The file this was in was less than 100 lines, so it wasn't excessively large.

PR #169, opened 2020-01-10
New Assignments

k3ntako:
New assignments will provide coach the ability to create categories (e.g., "Food", "Cars"). 
Members will be able to create assignments under the categories (e.g., "Salad" under "Food", "Jesse's Car" under "Cars").
Members will be able to sign up for these assignments created by other members and/or coach.

This PR only addresses the migration/models, and EventController methods and tests that broke as a result.

This PR will also have merge conflicts with the [mini PR](https://github.com/mighty-speed-krush/Mighty-Speed-Krush/pull/168). That PR would make the `EventController.update` simpler.

AssignmentController updates and tests are coming in next.


server/controllers/EventController/_listToDestroy.js
```diff
@@ -0,0 +1,33 @@
+const { Sequelize } = require('../../sequelize/models');
+
```
mienaikoe: I really think you're going overboard with separating everything. It's less readable to do this, and adds a bunch of code you didn't need to write that just shuffles params and fields around.

server/controllers/EventController/update.js
```diff
@@ -41,34 +23,51 @@ module.exports = async (req, res, next) => {
       throw new NotFoundError("Event with specified ID does not exist on this season");
     }
 
+    const { name, datetime, location_id, note, assignmentCategories } = validator(req.body);
```
mienaikoe: This should be done before database work

server/controllers/EventController/update.js
```diff
@@ -41,34 +23,51 @@ module.exports = async (req, res, next) => {
...
-            [Sequelize.Op.notIn]: assignmentIdsToUpdate,
-          },
-          event_id: req.params.event_id,
+      // Delete any assignment categories that are not in the request body
+      const destroyCategoryOptions = listToDestroy.categories(assignmentCategories, event_id, transaction);
+      await AssignmentCategory.destroy(destroyCategoryOptions);
+
+      for (const category of assignmentCategories) {
+        if (!category.name || !category.name.trim()) {
+          throw new ValidationError('Assignment category name is required');
```
mienaikoe: Assignment and Event->AssignmentCategory updates should be done in a separate endpoint.

server/sequelize/migrations/20191116170924-create-assignment.js
```diff
@@ -12,15 +12,28 @@ module.exports = {
         allowNull: false,
         type: Sequelize.STRING(255),
       },
+      description: {
+        allowNull: false,
+        type: Sequelize.STRING(750),
```
mienaikoe: TEXT might be better suited for a description

server/tests/EventControllerTests/update.js
```diff
@@ -164,10 +208,15 @@ module.exports = () => {
...
-            name: "Bring dogs",
-            assigneeLimit: 3,
+          assignmentCategories: [{
+            id: testAssignmentCategory.id,
+            name: "Food!",
+            assignments: [{
+              id: testAssignment.id,
+              name: "Edible gold leaf and diamonds",
+              assigneeLimit: 3,
+              description: "It might not be filling, but at least we will look super fancy",
```
mienaikoe: goals. Maybe one day we'll be that fancy

k3ntako:
Take out every assignment related thing in EventController except cascade delete

server/sequelize/migrations/20191116170924-create-assignment.js
k3ntako: I think we should have a front-end/controller limit. If so, what do you think is a good length? what do you think about 750 for now?

mienaikoe: yep that's fine.

k3ntako: I added a hook in Assignment that checks for assignment length

PR #171, opened 2020-01-11
SeasonController Tests

mienaikoe:


mienaikoe:
>This looks good, but I think I was confused about what active means. It could mean:

Yea this is something I was struggling with too but couldn't think of a better word.

> Maybe this is for a future PR, but do you think we should have API endpoints for delete and SeasonsForTeam?

Yes for delete. I think seasons for a team could come from the `Team.getMine()` request, but I see the benefit in fetching it separately. I'll handle deletion in this PR.

server/controllers/SeasonController/getMine.js
```diff
@@ -0,0 +1,21 @@
+const { Season, } = require('../../sequelize/models');
+const _fetchOptions = require('./_fetchOptions');
+
+
+module.exports = async function(req, res, next){
+  try{
+    const seasons = await Season.findAll(Object.assign({
+        where: {
+          team_id: req.session.team_id,
+          is_active: true
```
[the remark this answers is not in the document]
mienaikoe: I don't think we should add Op.and in, as sequelize has made its opinion that ANd is the default state. Also, getMine has been removed from this PR caus it's already implemented in SeasonPerson

server/tests/SeasonController/create.js
```diff
@@ -0,0 +1,58 @@
+const { expect } = require('chai')
+const SessionHelper = require('../helpers/SessionHelper');
+const test_async = require('../test_async');
```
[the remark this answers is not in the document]
mienaikoe: No. I don't want to rewrite the tests we already have, caus it would be too time consuming. Let's move forward with whatever you prefer, and get to the frontend soon.

server/controllers/SeasonController/getActive.js
```diff
@@ -1,25 +1,13 @@
...
 
 
 module.exports = async function(req, res, next){
   try{
-    if( !req.session.season_id ){
-      throw new ValidationError("You must specify a season_id");
-    }
-
     const season = await Season.findByPk(
       req.session.season_id,
```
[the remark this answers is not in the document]
mienaikoe: I like getSessionSeason.

server/controllers/SeasonController/_fetchOptions.js
```diff
@@ -0,0 +1,20 @@
+const { Person, SeasonPerson, Team } = require('../../sequelize/models');
+
+
+module.exports = {
+  attributes: ['id','name','email','is_active'],
+  include: [{
+    model: SeasonPerson,
+    attributes: ['person_id','season_id','role'],
+    include: {
```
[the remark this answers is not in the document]
mienaikoe: ah yep.

npm-debug.log
```diff
@@ -0,0 +1,49 @@
+0 info it worked if it ends with ok
```
k3ntako: guessing we don't want this?

server/controllers/SeasonController/delete.js
```diff
@@ -0,0 +1,50 @@
...
+    // season unless they made one accidentally
+    const season = await Season.findbyPk(req.params.season_id, {
+      include: [{
+        model: Event,
+        where: {
+          is_cancelled: false
+        },
+        required: true
+      }]
+    });
```
k3ntako: Don't we need to check that they are future events? Coach wouldn't cancel past events.

server/controllers/SeasonController/delete.js
```diff
@@ -0,0 +1,50 @@
...
+        "Seasons cannot be deleted with uncancelled events. Please cancel them and try again."
+      )
+    }
+
+    // delete the SeasonPeople, then the Season
+    await SeasonPerson.delete({
+      where: {
+        season_id: req.params.season_id
+      }
+    });
```
k3ntako: could we use `cascade` instead? or are you trying to be safe?

server/controllers/SeasonController/delete.js
```diff
@@ -0,0 +1,50 @@
...
+      throw new ValidationError(`no season found for id ${req.params.season_id}`)
+    }
+    if( season.Events.length > 0 ){
+      throw new ValidationError(
+        "Seasons cannot be deleted with uncancelled events. Please cancel them and try again."
+      )
+    }
+
+    // delete the SeasonPeople, then the Season
+    await SeasonPerson.delete({
```
k3ntako: I think the method is `destroy` not `delete`

server/controllers/SeasonController/delete.js
```diff
@@ -0,0 +1,50 @@
...
+      where: {
+        season_id: req.params.season_id
+      }
+    });
+    await Season.delete({
```
k3ntako: same as above, I think the method is `destroy` not `delete`

server/controllers/SeasonController/delete.js
mienaikoe: ah good point

server/controllers/SeasonController/delete.js
mienaikoe: Yea. idk I get weird about cascade caus Bloomberg always forbid its use.

server/controllers/SeasonController/delete.js
mienaikoe: yep. not sure how that worked

PR #172, opened 2020-01-12
SeasonPersonController Tests

mienaikoe:


server/controllers/SeasonPersonController/delete.js
```diff
@@ -1,24 +1,63 @@
...
+  await AssignmentPerson.destroy({
+    where: {
+      person_id,
+    },
+    include: [{
+      model: Assignment,
+      required: true,
+      include: _eventInclude(season_id)
+    }]
+  })
```
k3ntako: does this `include` approach work for `destroy`? It didn't seem to work for me, and I'm not sure why Sequelize would care about `include` for something they're destroying.

If it works for you, do you think you could write a test for it? We wouldn't want people to destroy anyone's SeasonPerson.

server/controllers/SeasonPersonController/update.js
```diff
@@ -1,10 +1,15 @@
...
+  sequelize, 
+  Sequelize 
+} = require('../../sequelize/models');
 
 
 module.exports = async function setRole(req, res, next){
   try{
     const role = req.params.role;
-    if( !SeasonPerson.rawAttributes.roles.values.includes(role) ){
+    if( SeasonPerson.rawAttributes.role.values.indexOf(role) === -1 ){
```
k3ntako: NIT: could use `includes` instead of `indexOf`

server/tests/SeasonPersonController/create.js
```diff
@@ -0,0 +1,86 @@
...
+        expect(seasonPerson.role).to.equal("INVITEE");
+      } else {
+        expect(seasonPerson.Person.email).to.equal(params.emails[1]);
+        expect(seasonPerson.role).to.equal("MEMBER");
+      }
+    });
+  });
+
+  it('should create SeasonPersons in the database for the given emails', async function() {
+    for( let ix in response ){
```
k3ntako: NIT: same as above, could use `for...of`

server/controllers/SeasonPersonController/create.js
```diff
@@ -101,7 +119,7 @@ module.exports = async function(req, res, next){
     }
 
     // Verify Coachness
-    const coachMembership = season.memberships.find( membership => membership.person_id === req.session.person_id );
+    const coachMembership = season.SeasonPeople.find( membership => membership.person_id === req.session.person_id );
```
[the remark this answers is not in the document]
mienaikoe: This isn't just a check. I use coachMembership further down in the method where I send an email to the invitee saying the coach invited them.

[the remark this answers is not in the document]
mienaikoe: i'll remove the if check tho.

server/controllers/SeasonPersonController/create.js
```diff
@@ -113,28 +131,30 @@ module.exports = async function(req, res, next){
...
     // Construct Memberships
-    const memberships = await sequelize.transaction( async transaction => {
-      return await Promise.all(
-        emails.map(
-          createMembershipForEmail.bind(null, season, transaction)
-        )
-      );
+    const memberships = [];
+    sequelize.transaction( async transaction => {
+      for( let ix in emails ){
```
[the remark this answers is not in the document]
mienaikoe: `emails` is defined near line 127

server/controllers/SeasonPersonController/delete.js
mienaikoe: include is required because filters can sometimes be part of inner joins that are used for filtering. Sequelize seems to prefer this route over using subqueries, which I personally prefer.

mienaikoe: I added a test for it, and it has a lot of dependencies on models that you were working on and likely changed for your PRs (Boat, Event, Assignment). 

How do you want to set the order we go in for these PRs, given those changes may have merge conflicts?

PR #173, opened 2020-01-13
AssignmentController 

k3ntako:
Updated to work with new schema for assignments and assignment categories.

Todo: tests (separate PR)


server/controllers/AssignmentController/_assigner.js
```diff
@@ -15,22 +15,32 @@ const deleteReplacePerson = async (transaction, assignment_id, replace_person_id
   }
 }
 
-module.exports = async (assignment_id, person_id, replace_person_id) => {
+module.exports = async (seasonId, assignment_id, person_id, replace_person_id) => {
   await sequelize.transaction(async transaction => {
     if (!person_id) {
       throw new ValidationError("No person ID provided");
     }
```
mienaikoe: nit: This could be done outside of the transaction

server/controllers/AssignmentController/create.js
```diff
@@ -18,13 +35,15 @@ module.exports = async (req, res, next) => {
 
     let assignee_limit = req.body.assigneeLimit;
     if (!Number.isInteger(assignee_limit) || assignee_limit < 1){
-      throw ValidationError('Assignee limit must be an integer greater than 0')
+      throw ValidationError('Assignee limit must be an integer greater than 0');
     }
```
mienaikoe: nit: do this validation at the beginning of the request handler so you don't have to make extra db calls without reason.

mienaikoe: Remove this validation after other branch is emrged.

PR #174, opened 2020-01-14
Assignment Category Tests

k3ntako:


server/controllers/AssignmentCategoryController/clone.js
```diff
@@ -7,7 +7,7 @@ module.exports = async (req, res, next) => {
       throw new ValidationError('No assignment category ID specified');
     }
 
-    if (!req.body.newEventId) {
+    if (!req.body.targetEventId) {
       throw new ValidationError('No new event ID specified');
```
mienaikoe: might want to change the wording of this too. The id isn't for a new event, but an event to clone.

server/tests/AssignmentCategoryController/update.js
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    // create person on separate team and season
+    otherTeamPerson = await createPersonWithTeam("AssignmentCategoryController_update@wakebird.com");
+  });
+
+  after(async () => {
+    await testCategory.destroy();
+    await testEvent.destroy();
+  });
```
mienaikoe: should also destroy otherTeamPerson and the team/season it was made with

server/tests/AssignmentCategoryController/delete.js
```diff
@@ -0,0 +1,72 @@
...
+      event_id: testEvent.id,
+    });
+
+    // create person on separate team and season
+    otherTeamPerson = await createPersonWithTeam("AssignmentCategoryController_delete@wakebird.com");
+  });
+
+  after(async () => {
+    await testEvent.destroy(); // will delete any associated boats
+  });
```
mienaikoe: should clean up otherTeamPerson

server/tests/AssignmentCategoryController/create.js
```diff
@@ -0,0 +1,75 @@
...
+        params: {
+          event_id: testEvent.id,
+        }
+      },
+      assertions: async (response) => {
+        expect(response).to.have.all.keys([ 'id', 'name', 'event_id', 'EventId' ]);
+        expect(response.name).to.equal(categoryName);
+
+        // check database
+        const category = await AssignmentCategory.findOne({
```
mienaikoe: you can save this to a higher-level variable so cleaning up the model doesn't require you to query by the name of the category.

server/controllers/AssignmentCategoryController/clone.js
```diff
@@ -7,7 +7,7 @@ module.exports = async (req, res, next) => {
...
     }
```
mienaikoe: everything from here to the response should be done in a transaction. It saves some time on performance, and also makes your changes atomic so nothing else can sneak something in while you do them.

server/controllers/AssignmentCategoryController/clone.js
```diff
@@ -58,7 +56,14 @@ module.exports = async (req, res, next) => {
       });
```
mienaikoe: there might be a way you can combine the category creation with the assignment creation. I've seen it on stackoverflow a couple times at least.

PR #178, opened 2020-01-18
SessionController - Join Methods

k3ntako:


server/controllers/SessionController/join.js
```diff
@@ -1,17 +1,18 @@
...
-const mapJoinParams = ({ first_name, last_name, email, password, verification_token }) => {
-  const parsedEmail = email.trim().toLowerCase(); // all emails in Wakebird are case insensitive and stored in lower case
-  const pass_salt = new Date().getTime();
-  return { 
-    first_name, 
-    last_name, 
-    email: parsedEmail, 
-    password, 
+const cleanParams = ({ first_name, last_name, email, password, verification_token }) => {
+  const parsedEmail = email && email.trim().toLowerCase(); // all emails in Wakebird are case insensitive and stored in lower case
```
mienaikoe: You can use EmailUtils to parse+verify this in the same step

server/controllers/SessionController/join.js
```diff
@@ -138,27 +144,44 @@ module.exports = async (req, res, next) => {
...
-      person = await createPerson(params);
+      const verification_token = await createPerson(params);
+      sendFirstTimeJoinEmail(params, verification_token);
     } else {
-      person = await createPersonFromInvite(req, person, personAuth, params);
+      await sequelize.transaction(async transaction => {
+        const personAuth = await upsertPersonAuth(params, transaction);
+
+        const { verification_token } = params;
+        const shouldVerify = verifyEmailToken(personAuth, verification_token);
```
mienaikoe: I'm not sure `shouldVerify` is the right word. It's more like `isLegit` or something

server/controllers/SessionController/join.js
```diff
@@ -31,104 +32,109 @@ const createPerson = async (params) => {
...
-  }).then(teams => {
-    // Transaction has been committed
-    if (teams.length === 1) {
-      req.session.team_id = teams[0].id;
-      req.session.is_coach = false;
-    }
-  }).catch(err => {
-    // Transaction has been rolled back
-    throw new Error(err);
+const updateRole = async (person, transaction) => {
```
mienaikoe: nit: this should be plural because it has the potential to modify multiple rows. Something more like `invitesToMembers`

server/controllers/SessionController/verifyEmail.js
```diff
@@ -18,9 +18,9 @@ module.exports = async (req, res, next) => {
     }
 
     let person = await Person.
-      findOne({ 
+      findOne({
         where: { email },
-        include: [{ model: Team, as: "teams" }] 
+        include: [{ model: Team, as: "teams" }]
```
mienaikoe: I think we should remove the `as` wherever we can.

server/sequelize/migrations/20191007232453-create-person-auth.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
         type: Sequelize.STRING
       },
       pass_salt: {
-        type: Sequelize.INTEGER
+        type: Sequelize.STRING(128)
```
mienaikoe: We'll have to remember to translate the current salts we have to hex

mienaikoe: Wanted to clarify. Because of this, I would rather keep salt as an integer.

k3ntako: We decided to keep this because we are converting the salt to string anyway in AuthUtils. All we have to do is convert the integer to string in the database.

server/tests/SessionController/resendVerification.js
```diff
@@ -0,0 +1,119 @@
...
+      expect(personAuth.verification_salt).to.not.equal("test");
+      expect(personAuth.verification_token_expiration.getTime()).to.be.greaterThan(Date.now());
+    });
+  });
+
+
+  describe('Attempt verify with invalid token', async () => {
+    after(() => sinon.restore());
+
+    it('should mark person as verified and remove their verification token info', function (done) {
```
mienaikoe: copypasta

server/tests/SessionController/resendVerification.js
```diff
@@ -0,0 +1,119 @@
+const { expect } = require('chai');
+const test = require('../test');
+const test_async = require('../test_async');
+const sinon = require('sinon');
+const SessionController = require('../../controllers/SessionController');
+const EmailUtils = require('../../utilities/EmailUtils');
+const SessionHelper = require('../helpers/SessionHelper');
+const { PersonAuth } = require('../../sequelize/models');
+
+let verifiedMember, unverifiedMember;
```
mienaikoe: clarify that these are sessions

server/controllers/SessionController/join.js
```diff
@@ -158,12 +158,12 @@ module.exports = async (req, res, next) => {
         const personAuth = await upsertPersonAuth(params, transaction);
 
         const { verification_token } = params;
-        const shouldVerify = verifyEmailToken(personAuth, verification_token);
+        const isLegitimate = verifyEmailToken(personAuth, verification_token);
 
-        if (shouldVerify) removeVerificationToken(personAuth, transaction);
+        if (isLegitimate) removeVerificationToken(personAuth, transaction);
```
mienaikoe: nit: braces on if statements

PR #181, opened 2020-01-23
Tests - Change Password

mienaikoe:


server/tests/SessionController/resetPassword.js
```diff
@@ -0,0 +1,170 @@
...
+          pass_salt: passSalt,
+          reset_token_hash: resetTokenHash,
+          reset_token_expiration: new Date(new Date().getTime() + (1000 * 60))
+        },
+      }, {
+        include: {
+          model: PersonAuth
+        }
+      });
+      console.log(testPerson);
```
k3ntako: NIT: probably don't need the `console.log`

server/tests/SessionController/updatePassword.js
```diff
@@ -0,0 +1,86 @@
...
+    });
+  });
+
+  after(async () => {
+    testPersonAuth && await testPersonAuth.destroy();
+    testSession && await testSession.destroy();
+  });
+
+
+  describe('Valid user can reset their password', async () => {
```
k3ntako: I think it should be "can **update** their password"

server/tests/SessionController/updatePassword.js
```diff
@@ -0,0 +1,86 @@
...
+          }
+        },
+      });
+    })
+
+    it('should return okay', async function () {
+      expect(response.okay).to.be.true;
+    });
+
+    it('should add password fields to associated PersonAuth', async function () {
```
k3ntako: do you think "add" is the correct word? Add suggests to me that the fields didn't already exist. I think "update" might be a better word.

server/tests/SessionController/updatePassword.js
```diff
@@ -0,0 +1,86 @@
...
+      const personAuth = await PersonAuth.findByPk(testPersonAuth.id);
+      
+      expect(personAuth).to.exist;
+      expect(personAuth.pass_salt).to.exist;
```
k3ntako: NIT: this is more of a question. You think there is any value in testing the length of the salt?

server/tests/SessionController/updatePassword.js
```diff
@@ -0,0 +1,86 @@
...
+          }
+        },
+        expectToThrow: true,
+        assertions: async (response, error) => {
+          expect(error).to.be.instanceOf(ForbiddenError);
+        },
+        done,
+      });
+    });
+  });
```
k3ntako: Same as above, don't think it's necessary, but we could test our password length requirement.

server/controllers/SessionController/updatePassword.js
```diff
@@ -1,18 +1,22 @@
...
-    let calculatedPasshash = passhashify(req.body.oldPassword, auth.salt);
-    if (calculatedPasshash !== auth.passhash) {
+    let calculatedPasshash = passhashify(req.body.old_password, auth.pass_salt);
+    if (calculatedPasshash !== auth.pass_hash) {
       throw new ForbiddenError("Password Did Not Match");
     }
 
-    const salt = new Date().getTime();
+    const salt = new Date().getTime().toString();
     const passhash = passhashify(req.body.password, salt);
```
k3ntako: I think we should check the password length

server/tests/SessionController/resetPassword.js
```diff
@@ -0,0 +1,170 @@
...
+        },
+        expectToThrow: true,
+        assertions: async (response, error) => {
+          expect(error).to.be.instanceOf(ForbiddenError);
+        },
+        done,
+      });
+    });
+  });
+}
```
k3ntako: Don't think it's necessary, but we could test our password length requirement.

PR #182, opened 2020-01-23
SessionController - me

mienaikoe:


server/controllers/SessionController/me.js
```diff
@@ -1,30 +1,51 @@
...
     );
 
     if (!person) {
       req.session.person_id = null;
-      throw new NotFoundError();
+      req.session.season_id = null;
+      req.session.team_id = null;
+      req.session.is_coach = null;
+      req.session.is_owner = null;
+      throw new NotFoundError("Your session has expired. Please login again.");
```
k3ntako: If person doesn't exist, doesn't that suggest that they don't even have an account? If so, I think the error message needs an update

server/tests/SeasonController/setActiveForTeam.js
```diff
@@ -33,6 +34,14 @@ module.exports  = () => {
...
     await testSession.destroy();
     await secondSeason.destroy();
+    const originalSeason = getGlobalSeason();
+    await Season.update({
+      is_active: true
+    }, {
+      where: {
+        id: originalSeason.id
+      }
+    });
```
k3ntako: NIT: I think it's easier to read if you did it this way:
```
originalSeason.is_active = true;
await originalSeason.save();
```

server/tests/SessionController/me.js
```diff
@@ -0,0 +1,76 @@
...
+    it('should return a valid me', async () => {
+      expect(response).to.exist;
+      expect(response).to.have.keys(FIELDS)
+      for( field in FIELDS ){
+        expect(response[field]).to.equal(testSession.person[field]);
+      }
+      expect(response.SeasonPeople.length).to.equal(1);
+      expect(response.SeasonPeople[0]).to.exist;
+      expect(response.SeasonPeople[0].Season).to.exist;
+      expect(response.SeasonPeople[0].Season.Team).to.exist;
```
k3ntako: do you think we should be more specific? Like check the IDs for season and team?

server/tests/SeasonController/setActiveForTeam.js
mienaikoe: is bug in sequelize

